PR #54, opened 2019-11-26
Fix classification example

Jean-Francois-Lafleche:
### Fix ModelNet
- Can use either ModelNet10 or ModelNet40
- Use arbitrary transforms rather than specific "rep" argument
- Fix ModelNet_Voxels class

### Fix classification example
- Fix incorrect imports
- Minor tweaks to improve readability
- Add Jupyter Notebook version with visualization

### Tests
OS: Ubuntu 18.04
GPU: RTX 2018Ti
CUDA: 10.1
Driver: 435.21

Commands:
- `pytest tests/`
- Jupyter Notebook run all for point_cloud_classifcication.ipynb


kaolin/conversions/meshconversions.py
```diff
@@ -78,7 +79,7 @@ def trianglemesh_to_voxelgrid(mesh: kaolin.rep.Mesh, resolution: int,
     if normalize:
         verts_max = mesh.vertices.max()
         verts_min = mesh.vertices.min()
-        mesh.vertices = (mesh.vertices - verts_min) / (verts_max - verts_min)
+        mesh.vertices = (mesh.vertices - verts_min) / (verts_max - verts_min) - 0.5
```
Caenorst, line 82: Can you explain this modification ?

kaolin/datasets/modelnet.py
```diff
@@ -12,28 +12,22 @@
...
+from kaolin.transforms import transforms as tfs
 
 
-_MODELNET10_CLASSES = ['bathtub', 'bed', 'chair', 'desk', 'dresser', 'monitor',
-                       'night_stand', 'sofa', 'table', 'toilet']
-
-
-class ModelNet10(torch.utils.data.Dataset):
-    r"""Dataset class for the ModelNet10 dataset.
+class ModelNet(object):
```
Caenorst, line 26: I believe ModelNet is now compatible with both ModelNet10 and ModelNet40. But ligne 30 is still `basedir (str): Path to the base directory of the ModelNet10 dataset.`

kaolin/datasets/modelnet.py
```diff
@@ -47,143 +41,129 @@ class ModelNet10(torch.utils.data.Dataset):
...
-        for category in categories:
-            assert category in all_classes, 'object class {0} not in list of availible classes: {1}'.format(
-                category, all_classes)
-            if train:
-                self.names += glob(
-                    root + '/ModelNet/volumetric_data/{0}/*/{1}/*{2}.mat'.format(category, 'train', side))
-            if test:
-                self.names += glob(
-                    root + '/ModelNet/volumetric_data/{0}/*/{1}/*{2}.mat'.format(category, 'test', side))
+    def __init__(self, basedir: str = 'data/ModelNet/', cache_dir: str = 'cache', train: bool = True, categories: list = ['bed'],
```
Caenorst: 1) I think `basedir` should be a required argument with no default value.
2) `basedir` don't have underscore but `cache_dir` does.

kaolin/datasets/modelnet.py
```diff
@@ -47,143 +41,129 @@ class ModelNet10(torch.utils.data.Dataset):
...
+        self.names = mesh_dataset.names
+        self.categories = mesh_dataset.categories
+        self.cat_idxs = mesh_dataset.cat_idxs
+
+        for res in self.params['resolutions']:
+            self.cache_transforms[res] = tfs.CacheCompose([
+                tfs.TriangleMeshToVoxelGrid(res, normalize=True, vertex_offset=0.5),
+                tfs.FillVoxelGrid(thresh=0.5),
+                tfs.ExtractProjectOdmsFromVoxelGrid()
+            ], self.cache_dir)
```
Caenorst, line 145: so `cache_dir` by default is `cache` which is relative to the current directory, which can lead to bad surprise. I suggest to put `cache_dir = None` by default and if `cache_dir is None` then `self.cache_dir = os.path.join(self.basedir + 'cache')`.

kaolin/datasets/modelnet.py
```diff
@@ -47,143 +41,129 @@ class ModelNet10(torch.utils.data.Dataset):
...
-            assert category in all_classes, 'object class {0} not in list of availible classes: {1}'.format(
-                category, all_classes)
-            if train:
-                self.names += glob(
-                    root + '/ModelNet/volumetric_data/{0}/*/{1}/*{2}.mat'.format(category, 'train', side))
-            if test:
-                self.names += glob(
-                    root + '/ModelNet/volumetric_data/{0}/*/{1}/*{2}.mat'.format(category, 'test', side))
+    def __init__(self, basedir: str = 'data/ModelNet/', cache_dir: str = 'cache', train: bool = True, categories: list = ['bed'],
+                 resolutions: List[int] = [32], device: str = 'cpu'):
```
Caenorst: device is now an argument, have you tested on both `cpu` and `gpu` ? (tests/datasets/test_ModelNet.py are commented out)

Caenorst (commented):
Good works. Some comments to solved and we are good to go.

kaolin/conversions/meshconversions.py
Jean-Francois-Lafleche: The expected behaviour of the `normalize` argument is to center the vertices around the origin. This is now also clarified in the updated docstring. This restores previous behaviour expected by the examples.

kaolin/datasets/modelnet.py
Jean-Francois-Lafleche: Agreed this is not ideal. `cache_dir` follows the format of the ShapeNet datasets, while `basedir` is the argument name used by the examples. This will need to be corrected in the future to ensure consistency between all dataset classes.

kaolin/datasets/modelnet.py
Jean-Francois-Lafleche: Great suggestion, implemented.

kaolin/datasets/modelnet.py
Jean-Francois-Lafleche: Agreed. Test has been enabled and updated for completeness.

kaolin/datasets/modelnet.py
Jean-Francois-Lafleche: Agreed, fixed.

Caenorst (approved):
Thanks for the quick fixes, LGTM

Issue #52, opened 2019-11-26
ModuleNotFoundError: No module named 'kaolin.nnsearch' 


Shabayek:
I'm using pytorch:19.10-py3 docker image (docker pull nvcr.io/nvidia/pytorch:19.10-py3). I get the following error:
.......

Finished processing dependencies for kaolin==0.2.0+a76a004
root@d404b78867d9:/workspace/kaolin# python
Python 3.6.9 |Anaconda, Inc.| (default, Jul 30 2019, 19:07:31) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import kaolin as kal
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/workspace/kaolin/kaolin/__init__.py", line 17, in <module>
    from kaolin import conversions
  File "/workspace/kaolin/kaolin/conversions/__init__.py", line 1, in <module>
    from kaolin.conversions.meshconversions import *
  File "/workspace/kaolin/kaolin/conversions/meshconversions.py", line 22, in <module>
    from kaolin.metrics.point import directed_distance as directed_distance
  File "/workspace/kaolin/kaolin/metrics/__init__.py", line 2, in <module>
    from .point import *
  File "/workspace/kaolin/kaolin/metrics/point.py", line 16, in <module>
    from kaolin.nnsearch import nnsearch
ModuleNotFoundError: No module named 'kaolin.nnsearch'

stephendb:
Same error. I installed pytorch using conda on ubuntu 18.04 as per the instructions.

shersoni610:
I had the same issue but it seems to get resolved if I run the code outside the directory i.e. above
README,md folder.

Shabayek:
@shersoni610 : I tried the same in both 1) the docker image and 2) a conda environment, but I get the following error:
In the docker image:
.....................
Using /opt/conda/lib/python3.6/site-packages
Finished processing dependencies for kaolin==0.2.0+a76a004
No CUDA runtime is found, using CUDA_HOME='/usr/local/cuda'
Traceback (most recent call last):
  File "kaolin/setup.py", line 192, in <module>
    ext_modules=cythonize(ext_modules),
  File "/opt/conda/lib/python3.6/site-packages/Cython/Build/Dependencies.py", line 897, in cythonize
    aliases=aliases)
[...]

Issue #43, opened 2019-11-23
ModuleNotFoundError: No module named 'Cython'


nikogamulin:
Environment: Ubuntu 18.04
When running python setup.py install, I get the following error:

```

> Using /home/niko/workspace/kaolin/venv/lib/python3.6/site-packages
> Finished processing dependencies for kaolin==0.2.0+1accd3e
> Traceback (most recent call last):
>   File "setup.py", line 98, in <module>
>     import Cython
> ModuleNotFoundError: No module named 'Cython'

```
### Installation procedure:
```
virtualenv -p /usr/bin/python3.6 venv #create virtual environment
# activate virtual environment
pip install numpy
pip install torch torchvision
python setup.py install
```

The same error occurred in case of using conda instead of pip

bobiblazeski:
Did you tried upgrading cython via conda?

`$ conda install -c anaconda cython`

nikogamulin:
> Did you tried upgrading cython via conda?
> 
> `$ conda install -c anaconda cython`

Thanks, this solved the issue.

Issue #40, opened 2019-11-22
No module named 'kaolin.nnsearch'


jbartolozzi:
OS: Ubuntu 16.04.6
Nvidia Driver: 418.87
CUDA: 10.1

Steps to reproduce
```
conda create --name kaolin python=3.7
conda activate kaolin
conda install pytorch-gpu numpy
python setup.py install
python -c "import kaolin"
```
```
(kaolin) jbartolozzi@mldwl4028-l|kaolin|: python -c "import kaolin"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/jbartolozzi/kaolin/kaolin/__init__.py", line 17, in <module>
    from kaolin import conversions
  File "/home/jbartolozzi/kaolin/kaolin/conversions/__init__.py", line 1, in <module>
    from kaolin.conversions.meshconversions import *
  File "/home/jbartolozzi/kaolin/kaolin/conversions/meshconversions.py", line 22, in <module>
    from kaolin.metrics.point import directed_distance as directed_distance
  File "/home/jbartolozzi/kaolin/kaolin/metrics/__init__.py", line 2, in <module>
    from .point import *
  File "/home/jbartolozzi/kaolin/kaolin/metrics/point.py", line 16, in <module>
    from kaolin.nnsearch import nnsearch
ModuleNotFoundError: No module named 'kaolin.nnsearch'
```

nikogamulin:
I had a similar problem while running python in kaolin base directory. If I run outside the kaolin directory, eveything works fine.

[Check issue #27 ](https://github.com/NVIDIAGameWorks/kaolin/issues/27)

stephendb:
Unfortunately this makes the pytest's unusable since they seem to have hard coded relative paths to the root directory.

Jean-Francois-Lafleche:
This error is occurring because the setup didn't compile the Cython function kaolin.nnsearch. After running setup, you probably saw the error `ModuleNotFoundError: No module named 'Cython'`. This does not occur with Python 3.6. 

Edit: on one of my tests, installing Cython manually did successfully compile nnsearch but did not resolve the import issue when in the root directory. Switching to Python 3.6 seems to be the only solution for now.

We will investigate a more permanent fix. Thank you for bringing it up and posting repro steps!

christinezuzart:
OS: Windows 10
Cuda: 10.1
Python : 3.7.6
pytorch: 1.4.0

[...]

Issue #27, opened 2019-11-20
ModuleNotFoundError: No module named 'kaolin.nnsearch'


prateep03:
Hi,

 I am unable to compile on Linux. Please help.

OS: Debian 9
Python: 3.6.9
Cuda: 10.0
pytorch: 1.3.1

OldConcept:
I had the same problem on Win10
![Z (@5W)Z )E_TGX4@WRR624](https://user-images.githubusercontent.com/46376624/69231649-67220600-0bc4-11ea-87c0-42543ec0569f.png)

krrish94:
Need to step outside of the `kaolin` base directory when running this.

https://github.com/NVIDIAGameWorks/kaolin/issues/3#issue-523193927

PR #66, opened 2019-11-29
Fix setup and import errors from kaolin root directory

Jean-Francois-Lafleche:
### Description
- Follow Cython's recommended best practices by distributing the `.cpp` files and making Cython compilation optional (see https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#distributing-cython-modules).
- Cython compilation can be triggered by setting the following env variable: `export USE_CYTHON=1` before running the setup.
- Cleanup `setup.py`, remove second setup script, add long description, add URL, improve readability.
- Update README with optional instruction to compile extensions inplace to prevent import errors when importing from the kaolin root directory.
- Update README to clarify requirements and installation.
- Update setup.py to pick up the version number indicated in `kaolin/version.py`.
- Update the project's version number to `0.1.0`.

### Related Issues
#40, #52, #43, #27, #21, #15, #14, #9, #3 

### What's Fixed
- Fix Cython related issues, Cython is no longer required during setup.
- Import errors when importing from kaolin root directory after `python setup.py install`.

### Tests
OS: Ubuntu 16.04
GPU: RTX 2080Ti
Driver: 435.21
CUDA: 10.1
Python: 3.6, 3.7

Commands:
- `python setup.py install; cd examples; python -c "import kaolin as kal; print(kal.__version__)"`
(note, this command by itself produces an import issue when importing kaolin from the kaolin root directory)
- `python setup.py build_ext --inplace; python setup.py install; python -c "import kaolin as kal; print(kal.__version__)"`
(note, when testing between different python versions, ensure to delete the `build` directory between tests)


setup2.py
```diff
@@ -0,0 +1,166 @@
...
+    'flake8',
+]
+
+
+# if __name__ == '__main__':
+# build_deps()
+setup(
+    # Metadata
+    name='f',
+    version='1',
```
krrish94: Should be `0.1.0`, to be consistent with everything else?

setup2.py
```diff
@@ -0,0 +1,166 @@
...
+        #     'kaolin/cuda/triangle_distance_cuda.cu',
+        # ]),
+        # CUDAExtension('kaolin.cuda.mesh_intersection', [
+        #     'kaolin/cuda/mesh_intersection.cpp',
+        #     'kaolin/cuda/mesh_intersection_cuda.cu',
+        # ]),
+        # CUDAExtension('kaolin.graphics.nmr.cuda.rasterize_cuda', [
+        #     'kaolin/graphics/nmr/cuda/rasterize_cuda.cpp',
+        #     'kaolin/graphics/nmr/cuda/rasterize_cuda_kernel.cu',
+        # ]),
```
krrish94: These could be safely removed (?)

setup2.py
```diff
@@ -0,0 +1,166 @@
...
+            'kaolin.nnsearch',
+            sources=[
+                'kaolin/cython/nnsearch.pyx'
+            ],
+            language='c++',
+        ),
+    ],
+    include_dirs=[np.get_include()],
+    # cmdclass={'build_ext': BuildExtension}
+)
```
krrish94: Need a newline here

setup.py
```diff
@@ -10,7 +10,7 @@
 VERSION = '0.1.0'
 DESCRIPTION = 'Kaolin: A PyTorch library for accelerating 3D deep learning research'
 URL = 'https://github.com/NVIDIAGameWorks/kaolin'
-AUTHOR = ''
+AUTHOR = 'Nvidia'
```
krrish94: Should double-check whether we write `Nvidia` or `NVIDIA`

Jean-Francois-Lafleche: Based on here: https://nvidia.frontify.com/d/qZxUpjTCnt96/nvidia-logo-quickstart-guide, I believe you are correct. Good catch!

setup2.py
Jean-Francois-Lafleche: It is, you are looking at a deleted file.

setup2.py
Jean-Francois-Lafleche: setup2.py is deleted

setup2.py
Jean-Francois-Lafleche: setup2.py is deleted

kaolin/cython/mcubes.cpp
```diff
@@ -891,7 +877,7 @@ static const char *__pyx_f[] = {
   "type.pxd",
 };
 
-/* "../../.local/lib/python3.6/site-packages/Cython/Includes/numpy/__init__.pxd":776
+/* "../../../miniconda3/envs/temptemp37/lib/python3.7/site-packages/Cython-0.29.14-py3.7-linux-x86_64.egg/Cython/Includes/numpy/__init__.pxd":776
```
Caenorst: I see that there is a lot of those diff in your PR, can you explain what is it doing ? (It seems commented, right ?)

Caenorst (commented):
The description of the PR says that the test was:
`python setup.py build_ext --inplace; python setup.py install;`
Is `python setup.py build_ext --inplace` required to installed kaolin now ? If yes shouldn't it be described in the README ? What happens if you only do `python setup.py install` ?

kaolin/cython/mcubes.cpp
Jean-Francois-Lafleche: Hmm, these files are generated by Cython, so I didn't dig into them You're right, I'll go through and clean them up.

Caenorst: If the files are generated by Cython, should we actually put them in the repository ? Shouldn't `setup.py` generate them for the user ?

Jean-Francois-Lafleche:
> The description of the PR says that the test was:
> `python setup.py build_ext --inplace; python setup.py install;`
> Is `python setup.py build_ext --inplace` required to installed kaolin now ? If yes shouldn't it be described in the README ? What happens if you only do `python setup.py install` ?

`python setup.py install` will work by itself in that it will install kaolin to the user's current environment's site-packages. However, imports from the kaolin root dir will fail as described above. So while `python setup.py build_ext --inplace` is optional (as indicated in the README), there's no cost to doing this command first (the subsequent install does not have to rebuild the extensions) and it avoids the aforementioned import issue.

For development, the command to run is still simply `python setup.py develop`.

All that said, I'll update the description to note that I've tested doing the install step by itself as well.

kaolin/cython/mcubes.cpp
Jean-Francois-Lafleche: See the first point in the description.

krrish94:
LGTM

kaolin/cython/mcubes.cpp
Jean-Francois-Lafleche: Comments (original source code) are now omitted from the generated C files.

PR #67, opened 2019-11-29
Implementation of DIB-R in kaolin core

TommyX12:
## Implementation of DIB-R in kaolin core
- [x] Copy dib-r implementation from `example` to `kaolin.graphics`
- [x] Get initial example working
  (Tested on Ubuntu 16.04, CUDA 10.0, PyTorch 1.3.1)
- [x] Clean-up and code improvement
  - [x] Remove old DIB-R code in example


kaolin/graphics/dib_renderer/rasterizer/rasterizer.py
```diff
@@ -0,0 +1,242 @@
...
+
+from __future__ import print_function
+from __future__ import division
+
+import torch
+import torch.nn
+import torch.autograd
+from torch.autograd import Function
+
+import kaolin.graphics.dib_renderer.cuda.rasterizer as cuda_rasterizer
```
Caenorst: This line makes my `import kaolin` fails, please use:
`from ..cuda import rasterizer as cuda_rasterizer`
or (but I would prefer relative import)
`from kaolin.graphics.dib_renderer.cuda import rasterizer as cuda_rasterizer`

examples/renderers/DIB-R/.gitignore
```diff
@@ -0,0 +1 @@
+results/
```
Caenorst: Not a big fan of multiple .gitignore can you put this in the root .gitignore ?

kaolin/graphics/dib_renderer/utils/mesh.py
```diff
@@ -365,51 +454,50 @@ def saveobjscale(meshfile, scale, maxratio, shift=None):
...
+
     meshfp.close()
     meshfp2.close()
-    
+
     return
 
 
-################################################################3    
+# 3
```
Caenorst: What is this commented out 3 ?

Caenorst (changes requested):
Thanks for the work, please fix the issues in the review and we are good to go :)

kaolin/graphics/dib_renderer/utils/mesh.py
TommyX12: My guess was that the original code had 3 as a typo ("#" is shift+3), and the auto formatter removed all extra "#"s. Nice catch :D removed.

kaolin/graphics/dib_renderer/rasterizer/rasterizer.py
TommyX12: Done; Also fixed other relative imports

examples/renderers/DIB-R/.gitignore
TommyX12: Done.

Caenorst (approved):
LGTM ! Thanks for the PR !

PR #74, opened 2019-12-06
Add Dockerfile

Caenorst:
Just add a Dockerfile to have a docker image of kaolin for CI tests / easy usage

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM nvcr.io/nvidia/pytorch:19.11-py3
+
+WORKDIR /kaolin
+COPY . .
+
+ENV KAOLIN_HOME "/kaolin"
+
+RUN pip install -r requirements.txt && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM nvcr.io/nvidia/pytorch:19.11-py3
+
+WORKDIR /kaolin
+COPY . .
+
+ENV KAOLIN_HOME "/kaolin"
+
+RUN pip install -r requirements.txt && \
+    python setup.py install
```
Jean-Francois-Lafleche, line 9: Should we not include the kaolin setup as part of the dockerfile?

Jean-Francois-Lafleche (approved):

Dockerfile
Caenorst: I don't understand your question, what `kaolin setup` ?

PR #77, opened 2019-12-07
Setup readthedocs

Jean-Francois-Lafleche:
### Description
Add support for readthedocs documentation builds

### Tests
OS: Ubuntu 18.04
Driver: 440.36
CUDA: 10.2
Python 3.6

Readthedocs - Confirmed documentations are successfully built from fork branch.

Commands:
- pytest tests/

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,9 @@
+version: 2
+
+python:
+   version: 3.7
+   system_packages: true
+   install:
+      - requirements: docs/requirements.txt
+      - method: setuptools
+        path: '.'
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,7 @@
 version: 2
 
 python:
-   version: 3.7
+   version: 3.6
    system_packages: true
    install:
       - requirements: docs/requirements.txt
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,6 +12,16 @@
 
 import os
 import sys
+
+
+# Get version number
+version_path = os.path.join('../', 'kaolin', 'version.py')
+with open(version_path, 'r') as f:
+    for row in f:
+        if '__version__' in row:
+            kal_version = row.split("'")[-2]
+            break
+
 sys.path.append('../')
 sys.path.append(os.path.abspath(os.path.join(
     os.path.dirname(__file__), '..', 'kaolin')))
@@ -25,10 +35,10 @@ master_doc = 'index'
 
 project = 'kaolin'
 copyright = '2019, NVIDIA Development Inc.'
-author = 'Krishna Murthy, Edward Smith, Wenzheng Chen, Amlan Kar, Jun Gao, Huan Ling, Clement Fuji Tsang, Ankur Handa, Sanja Fidler'
-
+author = 'NVIDIA'
+version = kal_version
 # The full version, including alpha/beta/rc tags
-release = '0.1.0 alpha'
+release = kal_version
 
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -94,7 +94,7 @@ def read(*names, **kwargs):
     )
 
     import torch
-    from torch.utils.cpp_extension import BuildExtension, CUDAExtension
+    from torch.utils.cpp_extension import BuildExtension, CUDAExtension, CUDA_HOME
```
krrish94: Curious as to what this import does.

setup.py
```diff
@@ -136,43 +136,47 @@ def read(*names, **kwargs):
...
-            'kaolin/graphics/nmr/cuda/rasterize_cuda_kernel.cu',
-        ]),
         triangle_hash_module,
         mise_module,
         mcubes_module,
         nnsearch_module,
     ]
+    
+    # If building with readthedocs, don't compile CUDA extensions
+    if os.getenv('READTHEDOCS') != 'True':
```
krrish94, line 146: Is this the recommended way to check if we're building on readthedocs?

krrish94 (commented):
CUDA_HOME appears to be unused.

And, I'm not 100% sure if the check for an envvar `READTHEDOCS` is the recommended way.

setup.py
Jean-Francois-Lafleche: Ah, you're right. I can take that out now. I used it in an earlier version when checking for whether the local machine has CUDA installed and setup properly. It tries to find the CUDA install path: https://pytorch.org/docs/stable/_modules/torch/utils/cpp_extension.html#CppExtension (see _find_cuda_home()

setup.py
Jean-Francois-Lafleche: Not sure if there is a recommended way but happy to change it if you have a alternative suggestion. This works well and can be easily extended with a CUDA check once we support CPU-only installation.

Jean-Francois-Lafleche: Ah, it appears this is the recommended way afterall: https://docs.readthedocs.io/en/stable/faq.html#how-do-i-change-behavior-for-read-the-docs

.readthedocs.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+
+python:
+   version: 3.7
```
Caenorst: Can we use python 3.6 as we have problems with 3.7 on kaolin ? It avoid us to have multiple python versions.

docs/conf.py
```diff
@@ -12,6 +12,16 @@
 
 import os
 import sys
+
+
+# Get version number
+version_path = os.path.join('../', 'kaolin', 'version.py')
```
Caenorst: why not just `version_path = '../kaolin/version.py' ?

docs/requirements.txt
```diff
@@ -0,0 +1,6 @@
+-f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+numpy
+torch_nightly
```
Caenorst: Do we really want to have dependency on torch_nightly ? Shouldn't we agree on a release instead ?

Caenorst (commented):
Just few questions to answer, otherwise LGTM

docs/requirements.txt
Jean-Francois-Lafleche: Fixed. Now using pinned pytorch version 1.3.1 with corresponding torchvision version.

docs/conf.py
Jean-Francois-Lafleche: Fixed. You are absolutely correct.

.readthedocs.yml
Jean-Francois-Lafleche: Fixed. Python 3.7 issues should be resolved with PR #66, but fixing to python 3.6 in any case as it is the development version and is more regularly tested.

PR #100, opened 2019-12-18
[WIP] Revert to old point loading behaviour

Jean-Francois-Lafleche:
### WIP
Confirming best path forward...

### Description
Minimum points sampled from mesh set to 100,000, which is then reduced in __get_item__ to the user-specified num_points arg. This produces significantly better results when training the GEOmetrics example.

### Tests
Trained GEOmetrics for 5 epochs
`pytest tests/datasets/test_ShapeNet.py`


Caenorst:
LGTM, is this what the publication is advising to do ?

Jean-Francois-Lafleche:
Old behaviour does not appear to offer benefits as originally thought.

Issue #97, opened 2019-12-16
Textures in DIB-R


tejank10:
Hello there! I was trying out the DIB-R example. I want to render it with texture file. I could understand how texture can be use in NMR and SoftRas but it is not clear in case of DIB-R example. `color` is not an optional argument for DIB-R renderer. In that case, how do I pass the colors of texture? I am having `.mtl` file for material and and an image for texture.

tejank10:
I also tried rendering using the approach described [here](https://github.com/NVIDIAGameWorks/kaolin/blob/master/kaolin/graphics/dib_renderer/test/test_tex.py). That led to:

```
Traceback (most recent call last):
  File "example.py", line 114, in <module>
    main()
  File "example.py", line 106, in main
    predictions, _, _ = renderer(points=[vertices, faces[0].long()], colors=[uvs, face_textures[0], textures])
  File "/home/tejan/.conda/envs/kaolin_new/lib/python3.6/site-packages/torch/nn/modules/module.py", line 541, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/tejan/.conda/envs/kaolin_new/lib/python3.6/site-packages/kaolin-0.1.0-py3.6-linux-x86_64.egg/kaolin/graphics/dib_renderer/renderer/base.py", line 65, in forward
  File "/home/tejan/.conda/envs/kaolin_new/lib/python3.6/site-packages/torch/nn/modules/module.py", line 541, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/tejan/.conda/envs/kaolin_new/lib/python3.6/site-packages/kaolin-0.1.0-py3.6-linux-x86_64.egg/kaolin/graphics/dib_renderer/renderer/texrender.py", line 77, in forward
  File "/home/tejan/.conda/envs/kaolin_new/lib/python3.6/site-packages/kaolin-0.1.0-py3.6-linux-x86_64.egg/kaolin/graphics/dib_renderer/rasterizer/rasterizer.py", line 101, in forward

RuntimeError: CUDA error: device-side assert triggered
The above operation failed in interpreter, with the following stack trace:
at <string>:3:9

def mul(a : float, b : Tensor) -> Tensor:
  return b * a
         ~~~~~ <--- HERE
def add(a : float, b : Tensor) -> Tensor:
  return b + a
def ne(a : float, b : Tensor) -> Tensor:
  return b != a
def eq(a : float, b : Tensor) -> Tensor:
  return b == a
[...]

PR #115, opened 2020-01-07
Refactored DIB-R and added textured DIB-R example

TommyX12:
Improved dib-r example (in `examples/renderers/DIB-R`) to be able to render textured mesh (#97). Refactored dib-r to be more intuitive for texturing.

Tested on Ubuntu 16.04, CUDA 10.0, PyTorch 1.3.1.


kaolin/graphics/dib_renderer/renderer/phongrender.py
```diff
@@ -83,10 +86,10 @@ def forward(self, points, cameras, colors, lightdirect_bx3, material_bx3x3, shin
         bnum = normal1_bxfx3.shape[0]
 
         # we have uv, normal, eye to interpolate
-        uv_bxpx2, ft_fx3, texture_bx3xthxtw = colors
-        c0 = uv_bxpx2[:, ft_fx3[:, 0], :]
-        c1 = uv_bxpx2[:, ft_fx3[:, 1], :]
-        c2 = uv_bxpx2[:, ft_fx3[:, 2], :]
+        # uv_bxpx2, texture_bx3xthxtw = colors
```
Caenorst: Please remove instead of comment it out

Caenorst (commented):
Can you specify the tests that you ran (and with what system) ? I see a bunch of modifications in the API of core functions, they seems reasonable but I'm afraid it could break something.

TommyX12:
> Can you specify the tests that you ran (and with what system) ? I see a bunch of modifications in the API of core functions, they seems reasonable but I'm afraid it could break something.

I believe there's currently no test in place for dib-r yet, and no other module is using the core API, so the change shouldn't affect anything.
I have also added the system info in the description (Ubuntu 16.04, CUDA 10.0, PyTorch 1.3.1).

kaolin/graphics/dib_renderer/renderer/phongrender.py
TommyX12: Resolved in latest commit.

Caenorst (approved):
LGTM

xiangdonglai:
Is 'ft_fx3' no longer supported? I think it is quite useful.

TommyX12:
> Is 'ft_fx3' no longer supported? I think it is quite useful.

`ft_fx3` is the face-to-vertex information. That information is already provided in the `points` argument, which should be a tuple of vertex position tensor and face information tensor. So it's not removed, it was just redundant. In the original DIB-R design, the same information needs to be passed to the method twice, so I removed one of them to reduce the chances of human error.

xiangdonglai:
There might be situations where ft_fx3 is not the same as faces_fx3. Sometimes ft_fx3 stores the indices of vertex texture coordinates ('uv' also known as 'vt'), which is different from face_fx3 that stores indices of vertices.

For example, on this [wiki page for obj](https://en.wikipedia.org/wiki/Wavefront_.obj_file), 'v1' are 'vt1' are different. v1,v2,v3 belong to faces_fx3, while vt1,vt2,vt3 belong to ft_fx3.
image

Besides, the util function that loads an obj file in kaolin also explicitly distinguishes between 'facenp_fx3' and 'ft_fx3'.
https://github.com/NVIDIAGameWorks/kaolin/blob/master/kaolin/graphics/dib_renderer/utils/mesh.py#L261

TommyX12:
> There might be situations where ft_fx3 is not the same as faces_fx3. Sometimes ft_fx3 stores the indices of vertex texture coordinates ('uv' also known as 'vt'), which is different from face_fx3 that stores indices of vertices.
> 
> For example, on this [wiki page for obj](https://en.wikipedia.org/wiki/Wavefront_.obj_file), 'v1' are 'vt1' are different. v1,v2,v3 belong to faces_fx3, while vt1,vt2,vt3 belong to ft_fx3.
> image
> 
> Besides, the util function that loads an obj file in kaolin also explicitly distinguishes between 'facenp_fx3' and 'ft_fx3'.
> https://github.com/NVIDIAGameWorks/kaolin/blob/master/kaolin/graphics/dib_renderer/utils/mesh.py#L261

That's a good point. I will add it back in a future PR.

xiangdonglai:
> > There might be situations where ft_fx3 is not the same as faces_fx3. Sometimes ft_fx3 stores the indices of vertex texture coordinates ('uv' also known as 'vt'), which is different from face_fx3 that stores indices of vertices.
> > For example, on this [wiki page for obj](https://en.wikipedia.org/wiki/Wavefront_.obj_file), 'v1' are 'vt1' are different. v1,v2,v3 belong to faces_fx3, while vt1,vt2,vt3 belong to ft_fx3.
> > image
> > Besides, the util function that loads an obj file in kaolin also explicitly distinguishes between 'facenp_fx3' and 'ft_fx3'.
> > https://github.com/NVIDIAGameWorks/kaolin/blob/master/kaolin/graphics/dib_renderer/utils/mesh.py#L261
> 
> That's a good point. I will add it back in a future PR.

That'll be great. Thanks.

PR #122, opened 2020-01-13
Use sparse array for VoxelGrid

Jean-Francois-Lafleche:
### Description
Improve write/read performance and reduce storage size by caching voxel grids as sparse arrays.

### Tests
OS: Ubuntu 18.04
CUDA: 10.2
Python 3.6
GPU: RTX 2080Ti
Driver: 440.44

Ran GEOmetrics example (includes multiple conversions to voxelgrid).


Caenorst (approved):
LGTM

PR #127, opened 2020-01-15
fix to graphics.softras.transform.Projection

cosw0t:


Caenorst:
Hi @michele-arrival !

Thank you for you interest in Kaolin ! Good catch !
We just updated CONTRIBUTING.md since #128 .

Please make sure that you are following the contribution guide, and that all of your commits are signed using git commit -s to signoff previous commits we recommend the method indicated [here](https://stackoverflow.com/questions/13043357/git-sign-off-previous-commits).

Also can you rebase your branch from the latest master ?

cosw0t:
Hi @Caenorst ! I've done as you suggested, thanks for looking into this

Edit: I've pushed the same merge request to the original repo https://github.com/ShichenLiu/SoftRas/ but i'm not sure when it will be look at. What is the strategy for kaolin? it seems redundant to have to push fixes in all repositories. How much is the code here going to diverge from the original implementation? Has the submodule approach been considered?

Caenorst:
Thanks @michele-arrival !

We used SoftRas code as a starting point, but we don't intend to keep track of it (unless we find an important bug fix / new feature).

We want to have a unified base class for all of our differentiable renderers (so you can change it without having to modify your scripts).

We also intend to optimize the renderers (memory and speed).

All of that may require us to diverge from the original implementation.

Caenorst (approved):
LGTM, thanks for the contribution ! :)

PR #131, opened 2020-01-16
Fix typo in DIB-R Phong Shader

avik-pal:
Signed-off-by: Avik Pal <avikpal@iitk.ac.in>


Caenorst (dismissed):
Good catch ! Thanks for your PR :)

Caenorst:
It seems that there is a problem with the signature. Can you please follow the steps [here](https://help.github.com/en/github/authenticating-to-github/checking-your-commit-and-tag-signature-verification-status) ?

Thanks again ! :)

Caenorst:
Hi @avik-pal ,

We removed the GPG verification for the signature as it is much constraint and replaced by a less restrictive verification, which ironically have some problems with your latest commit.

You can just remove the two latest commits and force push to your branch and it should be just fine

Thank you again!

Caenorst (approved):

PR #106, opened 2019-12-22
fix three_interpolate bug

ghost:
Currently, `three_interpolate_grad` calls `three_interpolate_kernel_launcher`, which is a bug. This PR fixes by calling `three_interpolate_grad_kernel_launcher` instead.

PR #133, opened 2020-01-18
fix three_interpolate bug

ghost:
Signed-off-by: Zheng Qin <clavichord93@hotmail.com>

Currently, `three_interpolate_grad` calls `three_interpolate_kernel_launcher`, which is a bug. This PR fixes this problem by calling `three_interpolate_grad_kernel_launcher` instead.

This PR is duplicated with #106 but is signed. Please close the old one.


Caenorst (approved):
LGTM, good catch and thanks for the fix ! :)

PR #141, opened 2020-01-29
Add trick to cross-compile over multiple cuda architectures in Dockerfile, setup.py and README.md

Caenorst:
Got it from: https://github.com/NVIDIA/apex/blob/master/setup.py#L9-L20


Jean-Francois-Lafleche (approved):
LGTM

PR #143, opened 2020-01-31
Slightly optimize loading and parsing obj file to mesh

Caenorst:
Optimizing the parsing of obj file to mesh, used in data loading.

(bonus) Also did a little optimization on the threshold function using in some data preprocessing. 

~ x1.8 speedup on loading ShapeNet kind of meshes

### DIFF
--- a/kaolin/conversions/voxelgridconversions.py
+++ b/kaolin/conversions/voxelgridconversions.py
@@ -336,10 +336,9 @@ def threshold(voxel: Union[torch.Tensor, VoxelGrid], thresh: float,
     """
     if isinstance(voxel, VoxelGrid):
         voxel = voxel.voxels
-        voxel = voxel.clone()
     helpers._assert_tensor(voxel)
-    if not inplace:
-        voxel[:] = voxel > tresh
+    if inplace:
+        voxel[:] = voxel > thresh
     else:
         voxel = (voxel > thresh).type(voxel.dtype)
     return voxel
--- a/kaolin/conversions/voxelgridconversions.py
+++ b/kaolin/conversions/voxelgridconversions.py
@@ -336,11 +336,12 @@ def threshold(voxel: Union[torch.Tensor, VoxelGrid], thresh: float,
     """
     if isinstance(voxel, VoxelGrid):
         voxel = voxel.voxels
-    if not inplace:
         voxel = voxel.clone()
     helpers._assert_tensor(voxel)
-    voxel[voxel <= thresh] = 0
-    voxel[voxel > thresh] = 1
+    if not inplace:
+        voxel[:] = voxel > tresh
+    else:
+        voxel = (voxel > thresh).type(voxel.dtype)
     return voxel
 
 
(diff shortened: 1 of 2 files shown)


kaolin/conversions/voxelgridconversions.py
```diff
@@ -336,11 +336,12 @@ def threshold(voxel: Union[torch.Tensor, VoxelGrid], thresh: float,
...
         voxel = voxel.voxels
-    if not inplace:
         voxel = voxel.clone()
     helpers._assert_tensor(voxel)
-    voxel[voxel <= thresh] = 0
-    voxel[voxel > thresh] = 1
+    if not inplace:
+        voxel[:] = voxel > tresh
+    else:
+        voxel = (voxel > thresh).type(voxel.dtype)
```
Jean-Francois-Lafleche, line 343: This does not result in an inplace operation. You're modifying the cloned tensor, same as above.

kaolin/conversions/voxelgridconversions.py
```diff
@@ -336,11 +336,12 @@ def threshold(voxel: Union[torch.Tensor, VoxelGrid], thresh: float,
     """
     if isinstance(voxel, VoxelGrid):
         voxel = voxel.voxels
-    if not inplace:
         voxel = voxel.clone()
```
Jean-Francois-Lafleche: The line `voxel = voxel.voxels` is pretty hard to read. The function may be more understandable by replacing the parameter name `voxel` with `voxelgrid`.

Jean-Francois-Lafleche (changes requested):
Looks good, just a couple comments regarding the conversion function.

kaolin/conversions/voxelgridconversions.py
Caenorst: then, shouldn't the attribute be named `voxel.voxelgrid` too ?

kaolin/conversions/voxelgridconversions.py
Caenorst: Good catch, fixed

kaolin/conversions/voxelgridconversions.py
Caenorst: the thing is that `voxel` is a `Union[torch.Tensor, VoxelGrid]`, which means it can be a voxelgrid or a voxel, which leads to this problem. Given that I'm not doing any modification on the naming part, I suggest that we fix that in another PR after a discussion.

Jean-Francois-Lafleche: My reasoning was that we have a `Voxelgrid` class which contains a `voxels` object, so `voxelgrid.voxels` follows that naming scheme. That said, I see your point and agree we can move this discussion to another PR.

kaolin/conversions/voxelgridconversions.py
Caenorst: Guess it's because I amended the commit but we can't see that I actually fixed the typo that I made. You can see it in the changes tho

Jean-Francois-Lafleche: Looks great!

Jean-Francois-Lafleche (approved):

PR #148, opened 2020-02-04
Cache using pickle

Jean-Francois-Lafleche:
### Description
Use Pytorch pickle functionality to cache data. In testing on caching tensors, pickling was 37.5% faster in write performance and 83.6% faster in read performance compared with the previous numpy caching function. Pickling also enables caching class instances such as PointClouds, VoxelGrids and Meshes.

### Code Origin
Code changes originate entirely from me.

### Tests
- **OS:** Ubuntu 18.04
- **GPU:** RTX 2080Ti
- **Driver:** 440.44
- **CUDA:** 10.2

#### Commands:
- pytest tests/

Signed-off-by: Jean-Francois Lafleche <jlafleche@nvidia.com>

### DIFF
--- a/kaolin/helpers.py
+++ b/kaolin/helpers.py
@@ -177,9 +177,9 @@ def _get_hash(x):
 
 
 class Cache(object):
-    """Caches the results of the called function to disk.
+    """Caches the results of a function to disk.
     If already cached, data is returned from disk, otherwise,
-    the function called is executed.
+    the function is executed. Output tensors are always on CPU device.
 
         Args:
             transforms (Iterable): List of transforms to compose.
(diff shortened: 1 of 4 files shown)


krrish94 (commented):
LGTM

Could we change the commit message to "Cache using pytorch" (as that's what this seems to be doing)?

kaolin/helpers.py
```diff
@@ -177,50 +177,54 @@ def _get_hash(x):
...
 
         Args:
             transforms (Iterable): List of transforms to compose.
             cache_dir (str): Directory where objects will be cached. Default
                              to 'cache'.
     """
 
-    def __init__(self, func: Callable, cache_dir: str = 'cache', cache_key: str = ''):
+    def __init__(self, func: Callable, cache_dir: [str, Path], cache_key: str):
+        assert isinstance(cache_dir, str) or isinstance(cache_dir, Path)
```
Caenorst: This should be written `isinstance(cache_dir, (str, Path))`.

I actually question the necessity of type enforcement for `cache_dir`, shouldn't we let the conversion to `Path` line 199 do the assertion for us ? (in case more types are supported)

An example from pytorch as reference: https://github.com/pytorch/vision/blob/master/torchvision/datasets/imagenet.py#L40

kaolin/helpers.py
```diff
@@ -177,50 +177,54 @@ def _get_hash(x):
...
+        assert isinstance(cache_key, str)
+        
+        # Assert cache dir and key are not empty
+        assert str(cache_dir).strip()
+        assert cache_key.strip()
```
Caenorst: `Path('') / ''` actually works, why do we absolutely want to avoid this case ? (especially since '.' would pass this assertion)

tests/test_helpers.py
```diff
@@ -0,0 +1,94 @@
...
+CACHE_DIR = 'tests/cache'
+
+
+@pytest.fixture(autouse=True)
+def cleanup():
+    """Cleanup after each test. """
+    yield
+    shutil.rmtree(CACHE_DIR)
+
+
```
Caenorst, line 21: Can we test also the correctness with a dictionary of mixed Tensor / numpy.array ?

Caenorst (changes requested):

kaolin/helpers.py
Jean-Francois-Lafleche: I agree, good catch.

kaolin/helpers.py
Jean-Francois-Lafleche: Agree path directory could be empty. However, the key should not since an empty key would likely only occur if the user forgot to pass it in and it would result in every item referring to the same data. I'd therefore leave it in just as a check, if you agree.

tests/test_helpers.py
Jean-Francois-Lafleche: Added a numpy array to the dictionary test.

kaolin/helpers.py
Caenorst: `cache_key` have no default value, so the user have to pass a value (which is good) otherwise calling the function would trigger a `TypeError: __init__() is missing 1 required positional argument ...`

If you want to exclude the `''` I think we should also exclude `'.'`, which would pass the current assertion, but would have the exact same effect.

Jean-Francois-Lafleche: Valid. I'll remove that check too then since it doesn't in fact improve usability. Thanks!

Jean-Francois-Lafleche: I removed the check to for a str type as well and cast to str when setting the `self.cache_dir` value. This will allow objects with a `__repr__` that can be used as a key to be passed in without error.

Caenorst (approved):
LGTM, thanks for the fixes !

PR #159, opened 2020-02-17
Vertex coordinate fix for obj loader

tovacinni:
Although I'm assuming Kaolin's obj loader (in kaolin.rep.Mesh.from_obj) only supports 2D textures, some obj files with 2D textures still include a zeroed-out 3rd dimension for the vertex texture coordinates. To support these files, I added a simple fix to set the extent of the slice. 

All tests passed but tests requiring ModelNet and ShapeNet were skipped.

Signed-off-by: tovacinni <tovacinni@gmail.com>


Caenorst (approved):
Thanks for the fix ! :)

PR #184, opened 2020-03-23
Add a base dataset that can used multiprocessed preprocessing.

Caenorst:
1) Add a base KaolinDataset that can use multiprocessed preprocessing.
2) Add a helper base dataset that combine multiple KaolinDataset
2) Implement an example of KaolinDataset usage for ShapeNet

The multiprocessed preprocessing allowed to great speedup (up to x8 on my machine for conversion into Voxels)

ShapeNet roughly tested with Mesh, Voxel and points.

**Test environment**
GPU: Titan RTX
OS: Ubunbut 18.04
docker image: built from Dockerfile

Comments:
this will probably require some fine-tuning and more thorough testing once KaolinDataset is extended to all datasets (Coming soon).


kaolin/datasets/base.py
```diff
@@ -0,0 +1,154 @@
+# Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
```
krrish94: 2020?

kaolin/datasets/base.py
```diff
@@ -0,0 +1,154 @@
...
+       3) _get_attributes:
+           Attributes getter that will be preprocess / transform independent.
+       4) __len__:
+           Return the size of the dataset
+    """
+    def __init__(self, *args, preprocessing_transform=None, preprocessing_params: dict = None,
+                 transform=None, no_progress: bool = False, **kwargs):
+        """
+        Args:
+            positional and keyword arguments for initialize(*args, **kwargs) (see class and initialize documentation)
```
krrish94, line 66: Is there a way we can "link" from here to the metaclass docs?

kaolin/datasets/base.py
```diff
@@ -0,0 +1,154 @@
...
+                - 'num_workers': number of process used in parallel for preprocessing (default: number of cores)
+            preprocessing_transform (Callable): Called on the outputs of _get_data over the indices
+                                                from 0 to len(self) during the construction of the dataset,
+                                                the preprocessed outputs are then cached to 'cache_dir'.
+            transform (Callable): Called on the preprocessed data at __getitem__.
+            no_progress (bool): disable tqdm progress bar for preprocessing.
+        """
+        self.initialize(*args, **kwargs)
+        if preprocessing_transform is not None:
+            desc = 'applying preprocessing'
```
krrish94, line 78: 'applying prepreocessing' -> 'preprocessing ...'?

kaolin/datasets/base.py
```diff
@@ -0,0 +1,154 @@
...
+    @abstractmethod
+    def _get_data(self, index):
+        pass
+
+    @abstractmethod
+    def __len__(self):
+        pass
+
+class CombinationDataset(KaolinDataset):
+    """Dataset combining multiple list of datasets outputs as inputs
```
krrish94: The description isn't super-clear imo

kaolin/datasets/base.py
```diff
@@ -0,0 +1,154 @@
...
+    the output of _get_data wiil be a tuple of all the 'data' of the __getitem__ of the dataset list
+
+    Args:
+        datasets: list or tuple of KaolinDataset
+    """
+    def _initialize(self, datasets):
+        self.len = len(datasets[0])
+        for i, d in enumerate(datasets):
+            assert len(d) == self.len, \
+                "dataset {} have different length ({})than the first dataset ({})".format(i, len(d), self.len)
```
krrish94: Can change the error message to be more descriptive
```
f"All datasets must have the same length. Invalid length at index {i} (expected: {self.len}, got: {len(d)})"
```
Note: used fstring only for convenience

kaolin/datasets/shapenet.py
```diff
@@ -17,6 +17,7 @@
 from pathlib import Path
 import torch
 import torch.utils.data as data
+from torch.multiprocessing import Pool
```
krrish94: Switch order of imports

kaolin/datasets/shapenet.py
```diff
@@ -126,15 +96,12 @@ class ShapeNet_Meshes(data.Dataset):
         train (bool): return the training set else the test set
```
krrish94: Return train set (if True), test set otherwise.

kaolin/datasets/shapenet.py
```diff
@@ -126,15 +96,12 @@ class ShapeNet_Meshes(data.Dataset):
...
         split (float): amount of dataset that is training out of 1
```
krrish94:
```
split (float): fraction of the dataset to be used for training (>=0 and <=1)
```

kaolin/datasets/shapenet.py
```diff
@@ -193,6 +160,85 @@ def __getitem__(self, index):
         return {'data': data, 'attributes': attributes}
 
 
+class ShapeNet(KaolinDataset):
+    r"""ShapeNet Dataset class for meshes.
+    Args:
+        root (str): path to ShapeNet root directory
```
krrish94, line 165: We should maybe specify the version of ShapeNet we're expecting

kaolin/datasets/shapenet.py
```diff
@@ -193,6 +160,85 @@ def __getitem__(self, index):
...
+        categories (list): List of categories to load from ShapeNet. This list may
+                           contain synset ids, class label names (for ShapeNetCore classes),
+                           or a combination of both.
+        train (bool): is using training split (default: True)
```
krrish94:
```
split to use (train if True, test otherwise) (default:True)
```

kaolin/datasets/shapenet.py
```diff
@@ -193,6 +160,85 @@ def __getitem__(self, index):
...
+        split (float): proportion of the dataset to split to the training datase
```
krrish94:
```
fraction of the dataset to be used for training (>=0 and <=1)
```

kaolin/datasets/shapenet.py
```diff
@@ -193,6 +160,85 @@ def __getitem__(self, index):
...
+    Returns:
+        .. code-block::
```
krrish94, line 172: Don't we need a blank line below `.. code-block::` in order for it to render properly when built?

kaolin/datasets/shapenet.py
```diff
@@ -193,6 +160,85 @@ def __getitem__(self, index):
...
+
+    def initialize(self, root: str, categories: list, train: bool = True, split: float = .7):
+        """Initialize the dataset
+           Args:
+               root (str): path to ShapeNet root directory
+               categories (list): List of categories to load from ShapeNet. This list may
+                                  contain synset ids, class label names (for ShapeNetCore classes),
+                                  or a combination of both.
+               train (bool): is using training split (default: True)
+               split (float): proportion of the dataset to split to the training dataset"""
```
krrish94: Similar comments as above for the docstrings.

krrish94 (changes requested):
A few docstring fixes, etc. LGTM otherwise.

kaolin/datasets/base.py
Caenorst: ? It's already 'applying preprocessing'

kaolin/datasets/base.py
Caenorst: Sure, can you suggest something ?

kaolin/datasets/base.py
krrish94: Yeah, I was suggesting changing to just `preprocessing ...`.

kaolin/datasets/base.py
Caenorst: It leads to indentations problems, also please notice that the docstrings are not exactly the same

kaolin/datasets/shapenet.py
Caenorst: ShapeNetV1 ?

kaolin/datasets/shapenet.py
Caenorst: Good catch, fixed (and I also fixed indentation)

kaolin/datasets/shapenet.py
Caenorst: Fixed

kaolin/datasets/shapenet.py
Caenorst: Fixed, I used [this](https://pytorch.org/docs/stable/torchvision/datasets.html) for reference

kaolin/datasets/base.py
Caenorst: That's better, done

kaolin/datasets/base.py
Caenorst: I think it's more descriptive to have:
```applying preprocessing: 100%|##################################| 75/75 [00:00<00:00, 1893.97it/s]```

File: kaolin/datasets/base.py (hunk 1 above)
Comment (Caenorst) [reply kaolin/datasets/base.py]: Good catch, fixed

File: kaolin/datasets/shapenet.py (hunk 13 above)
Comment (Caenorst) [reply kaolin/datasets/shapenet.py]: Fixed

File: kaolin/datasets/shapenet.py (hunk 7 above)
Comment (Caenorst) [reply kaolin/datasets/shapenet.py]: This is actually not part of my PR, we intend to destroy this class to replace it by the one I've made.

File: kaolin/datasets/shapenet.py (hunk 8 above)
Comment (Caenorst) [reply kaolin/datasets/shapenet.py]: This is actually not part of my PR, we intend to destroy this class to replace it by the one I've made.

File: kaolin/datasets/shapenet.py (hunk 6 above)
Comment (Caenorst) [reply kaolin/datasets/shapenet.py]: I removed it as it is unused

File: kaolin/datasets/base.py (hunk 2 above)
Comment (krrish94) [reply kaolin/datasets/base.py:66]: Ah I see. I was thinking of the way pytorch does it. Eg. https://pytorch.org/docs/stable/tensors.html#torch.Tensor.sigmoid

Not a concern atm, was just wondering.

File: kaolin/datasets/base.py (hunk 3 above)
Comment (krrish94) [reply kaolin/datasets/base.py:78]: `Preprocessing` is a verb in itself, so I'm leaning towards it. But this is fine too, if you prefer.

File: kaolin/datasets/base.py (hunk 4 above)
Comment (krrish94) [reply kaolin/datasets/base.py]: It's tricky to explin, yeah. How about something like
```
Combines a list of datasets into a unified dataset object. Useful when multiple output representations are needed from a common base representation (Eg. when a mesh is to be served as both a pointcloud and a voxelgrid, etc.)
```

File: kaolin/datasets/shapenet.py [hunk 14]
```diff
@@ -193,6 +160,85 @@ def __getitem__(self, index):
...
+
+        self.names = [p.name for p in self.paths]
+
+    def __len__(self):
+        """Returns the length of the dataset. """
+        return len(self.paths)
+
+    def _get_data(self, index):
+        synset_idx = self.synset_idxs[index]
+        obj_location = self.paths[index] / 'model.obj'
```
krrish94, line 228: Actually, we can support `ShapeNetCoreV1` as well as `ShapeNetCoreV2` if we check for both `model.obj` and `model_normalized.obj`. (V2 has a few categories renamed to `model_normalized.obj`). Can be done in another PR if you prefer.

kaolin/datasets/shapenet.py
krrish94: Yeah, and also see my other comment on `model.obj` and `model_normalized.obj`

krrish94 (changes requested):
Almost there, minor clarifications needed

kaolin/datasets/shapenet.py
Caenorst: The PR is actually not about ShapeNet, the class is more here to test the base dataset. Let's set it for ShapeNetV1

Caenorst:
I think all the comments have been resolved now

kaolin/datasets/shapenet.py
Caenorst: Done

kaolin/datasets/shapenet.py
krrish94: Agreed

krrish94:
LGTM

krrish94 (approved):

PR #187, opened 2020-03-27
Base dataset formatting and bug fixes

TommyX12:
- Fixed `KaolinDataset` where it incorrectly calls `_initialize` instead of `initialize`
- Changed behavior of transform to not automatically spread data returned from `_get_data`


Caenorst (approved):
LGTM, thanks for the fixes :)

PR #189, opened 2020-03-30
Refactored ModelNet dataset to use KaolinDataset

TommyX12:
- Fixed bugs in `KaolinDataset` base class
- Refactored `ModelNet` dataset to use `KaolinDataset`


kaolin/datasets/modelnet.py
```diff
@@ -12,63 +12,62 @@
...
 import torch
 import os
 from glob import glob
-from tqdm import tqdm
 
 from kaolin.rep.TriangleMesh import TriangleMesh
-from kaolin.transforms import transforms as tfs
 
 
 class ModelNet(object):
```
Caenorst: It's not using KaolinDataset as a base class

kaolin/datasets/modelnet.py
```diff
@@ -12,63 +12,62 @@
...
     """
 
-    def __init__(self, basedir: str,
-                 split: Optional[str] = 'train',
-                 categories: Optional[Iterable] = ['bed'],
-                 transform: Optional[Callable] = None,
-                 device: Optional[Union[torch.device, str]] = 'cpu'):
+    def initialize(self, root: str,
+                   split: Optional[str] = 'train',
+                   categories: Optional[Iterable] = ['bed']):
```
Caenorst: There shouldn't be any default value for categories.

On a side note, if you want to set a default argument value it shouldn't be a mutable value, it should be `None` and then the default value is set inside the function.

Caenorst (commented):
Can you remove the files that you modified from .flake8 `exclude` list ?

kaolin/datasets/modelnet.py
TommyX12: Done.

kaolin/datasets/modelnet.py
TommyX12: Done.

Caenorst: An update, can you set all categories by default ?

Caenorst:
Also just need to sign your commits, and you're good to go

Caenorst (approved):
LGTM

PR #190, opened 2020-03-31
Refactor SHREC16 to inherit from KaolinDataset

krrish94:


kaolin/datasets/shrec.py
```diff
@@ -35,70 +37,123 @@ class SHREC16(Dataset):
         categories (list): List of categories to load (each class is
             specified as a string, and must be a valid `SHREC16`
             category).
-        mode (str, choices=['train', 'test']): Whether to load the
-            'train' split or the 'test' split
+        train (optional, bool): If True, return the train split, else return the test
```
Caenorst: Remove "optional" from the type description (you already put a default value)

kaolin/datasets/shrec.py
```diff
@@ -35,70 +37,123 @@ class SHREC16(Dataset):
...
-
-        super(SHREC16, self).__init__()
-
-        if mode not in ['train', 'test']:
-            raise ValueError('Argument \'mode\' must be one of \'train\''
-                             'or \'test\'. Got {0} instead.'.format(mode))
+    def initialize(
+        self,
+        root: str,
+        categories: Optional[Iterable] = None,
```
Caenorst: I don't think categories should have a default value, it's not specified in the documentation neither

kaolin/datasets/shrec.py
```diff
@@ -35,70 +37,123 @@ class SHREC16(Dataset):
...
+        train: Optional[bool] = True,
```
Caenorst: It shouldn't be Optional if the default value is True, Optional make it handle None, but AFAIK it shouldn't handle anything else than True and False

kaolin/datasets/shrec.py
```diff
@@ -35,70 +37,123 @@ class SHREC16(Dataset):
...
-            'train' split or the 'test' split
+        train (optional, bool): If True, return the train split, else return the test
+            split (default: True).
     Returns:
-        dict: Dictionary with keys: 'vertices' : vertices , 'faces' : faces
+        .. code-block::
+
+           dict: {
+                attributes: {path: str, category: str, label: int},
+                data: {vertices: torch.Tensor, faces: torch.Tensor}
```
Caenorst: I believe data are a kaolin.rep.TriangleMesh now.

Also (not a strong request, feel free to ignore from now), we probably want to change the formation in the description, because those return type depend on preprocessing_transform and transform.

kaolin/datasets/shrec.py
krrish94: It is `optional` because we have a default value.

kaolin/datasets/shrec.py
krrish94: I think it makes sense to have a default value. Although I feel, my choice of the default isn't well-motivated. Let's have the default behavior to _load all images from the dataset_.

krrish94: Added to the docs as well.

kaolin/datasets/shrec.py
krrish94: Nice catch. For docstring descriptions, I'd like to do a later PR that spans all datasets.

kaolin/datasets/shrec.py
Caenorst: I think there is a misunderstand on the definition of "optional" the type `Optional` (usually used for type hints) means that it can be `None` (https://docs.python.org/3/library/typing.html#typing.Optional).

Having a default value is already indicated by `(default: True)`

kaolin/datasets/shrec.py
```diff
@@ -34,71 +36,125 @@ class SHREC16(Dataset):
...
-            raise ValueError('Argument \'mode\' must be one of \'train\''
-                             'or \'test\'. Got {0} instead.'.format(mode))
+    def initialize(
+        self,
+        root: str,
+        categories: Iterable = None,
+        train: bool = True,
+    ):
 
         VALID_CATEGORIES = [
```
Caenorst: I think this should be static (defined inside the class outside of the function)

Caenorst:
Can you remove the files that you corrected from the exclusion list in .flake8 config file ?

kaolin/datasets/shrec.py
krrish94: Yes

krrish94:
Good catch on the `flake8` ignore. Fixed

Caenorst (commented):
Can you add a test in `tests/datasets/` ?

Caenorst (approved):
LGTM, thanks for the contrib !

PR #194, opened 2020-04-07
Fix ODM example

Jean-Francois-Lafleche:
# Tests

Ubuntu 18.04
CUDA 10.1
Pytorch 1.4

- Train + eval of Direct mode (0.776 IoU) on 'chair' class (20 epochs)
- Train + eval of MVD mode (0.773 IoU) on 'chair' class (100 epochs)
- Resume training
- Visualization

# Notes

- IoU is good but results don't look as good as they do in paper.
- In above test, MVD mode did worse than Direct mode.


examples/SuperResolution/odm/train.py
```diff
@@ -0,0 +1,259 @@
...
+class Engine(object):
+    """Engine that runs training and inference.
+    Args
+        model_name (str): Name of model being trained. ['Direct', 'MVD']
+        cur_epoch (int): Current epoch.
+        print_every (int): How frequently (# batches) to print loss.
+        validate_every (int): How frequently (# epochs) to run validation.
+    """
+
+    def __init__(self, mode, print_every=1, resume_name=None):
```
Caenorst: the arguments doesn't correspond to the doc string

examples/SuperResolution/odm/train.py
```diff
@@ -0,0 +1,259 @@
...
+
+            pred_odms = self._get_pred(data)
+            loss = self._get_loss(pred_odms, data)
+
+            optimizer.zero_grad()
+            loss.backward()
+            loss_epoch += loss.item()
+
+            # logging
+            num_batches += 1
```
Caenorst: I don't see the difference between `num_batches` and `i + 1`, can we use just one variable ? (use `batch_idx` instead of `i`)

examples/SuperResolution/odm/utils.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def to_occupancy_map(inp, threshold=None):
+    if threshold is None: 
+        threshold = inp.shape[-1]
+    zeros = inp < threshold
+    ones = inp >= threshold
+    inp = inp.clone()
+    inp[ones] = 1 
+    inp[zeros] = 0
```
Caenorst: I believe the equivalent:
`inp = (inp >= threshold).type(inp.dtype)`
would be simpler and faster.

examples/SuperResolution/odm/train.py
```diff
@@ -0,0 +1,259 @@
...
+        if save_best:
+            torch.save(model.state_dict(), os.path.join(logdir, 'best.pth'))
+            torch.save(optimizer.state_dict(), os.path.join(logdir, 'best_optim.pth'))
+            # Log other data corresponding to the recent model
+            with open(os.path.join(logdir, 'best.log'), 'w') as f:
+                f.write(json.dumps(log_table))
+            tqdm.write('====== Overwrote best model ======>')
+
+    def _get_pred(self, data):
+        pred_fns = {
```
Caenorst: Why using a dictionary here and `if ... elif` branching line 231 ?
Maybe we could set all the branching at __init__ ? (self.mode shouldn't change during usage right ?)

examples/SuperResolution/odm/train.py
```diff
@@ -0,0 +1,259 @@
...
+            'Direct': self._get_pred_direct,
+            'MVD': self._get_pred_mvd,
+        }
+        return pred_fns[self.mode](data)
+
+    @staticmethod
+    def _get_pred_direct(data):
```
Caenorst: Why using a staticmethod ?

This function is using 2 global variables (`args.device` and `model`) which make it harder to understand, shouldn't those be attributes variables and then declare those functions (`_get_pred_direct` and `_get_pred_mvd`) as instance methods ?

examples/SuperResolution/odm/eval.py
```diff
@@ -0,0 +1,109 @@
...
+        inp_voxels = data[32]['voxels'].to(args.device)
+
+        # inference
+        if args.mode == 'Direct':
+            pred_odms = model(inp_odms)
+        elif args.mode == 'MVD':
+            initial_odms = upsample_odm(inp_odms) * 4
+            distance = 128 - initial_odms
+            pred_odms = model(inp_odms) * distance
+
```
Caenorst, line 78: I guess there is a missing comment here if you separate in two for the same branch condition, otherwise it could be merge in a single if / elif.

Caenorst (commented):
Thanks for the contribution!

Few comments:
`train.py` have a very different structure than `eval.py` while they should share a lot of features.
For instance:
- using `Engine` class in `train.py` but still using some global variable such as model and args.device vs `eval.py` being more procedural.
- directly transfer all the data to args.device at the beginning of each iteration in `eval.py`.

Overall I find `eval.py` style more readable (especially since it's a example script), but if you want to keep the Engine class maybe add some more comments, make Engine code more self-contained (move global variables to attributes), and convert to args.device the data at the very beginning of the code.

Finally in `.flake8` in the exclusion list can you replace `examples/SuperResolution` by `examples/SuperResolution/voxel-ModelNet`
`examples/SuperResolution/voxel-ShapeNet` ?

examples/SuperResolution/odm/utils.py
Jean-Francois-Lafleche: Thanks, I didn't look at fixing up these - I should have.

examples/SuperResolution/odm/train.py
Jean-Francois-Lafleche: Correct, this was an oversight.

examples/SuperResolution/odm/train.py
Jean-Francois-Lafleche: I 100% agree with the overall structure and style comments you have described. It was written this way for consistency with other examples. I have re-written the code to be easier to read and done away with unnecessary functions and classes.

examples/SuperResolution/odm/train.py
Jean-Francois-Lafleche: Removed all this unnecessary complexity.

examples/SuperResolution/odm/train.py
Jean-Francois-Lafleche: Removed all this unnecessary complexity.

examples/SuperResolution/odm/eval.py
Jean-Francois-Lafleche: Correct, comment missing - I split this into two steps to separate the inference from the IoU steps. IoU comment was missing.

Jean-Francois-Lafleche:
Currently re-training to confirm that no regression occurred after these latest fixes.

examples/SuperResolution/odm/preprocessor.py
```diff
@@ -0,0 +1,46 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from kaolin.transforms import transforms as tfs
+
+
+class VoxelODMs:
```
Caenorst: you can make it inherit from kaolin.Transforms ? (also you can remove `__repr__')

Jean-Francois-Lafleche: Yup!

shumash:

PR #195, opened 2020-04-07
Improve preprocessing perf when already cached

Jean-Francois-Lafleche:
# Summary

Only loop through files that have not already been cached. This greatly reduces time spent in preprocessing loop if some or all of the preprocessing had previously occurred.

# Tests

Ubuntu 18.04
CUDA 10.1
Pytorch 1.4

- Ran new ODM example with VoxelODM preprocessor.


Caenorst (approved):
LGTM, nice feature ! :)

PR #196, opened 2020-04-07
Fix OccNet example

Jean-Francois-Lafleche:
Update OccNet example.

**Tests**

Ubuntu 18.04
CUDA 10.1
Pytorch 1.4

- Trained for 280 epochs on the category 'chair' with Blender renderings.
- Resumed training.
- Ran evaluation and obtained IoU of 0.4675
- Ran pytest


examples/ImageRecon/OccNet/eval.py
```diff
@@ -12,135 +12,139 @@
...
+                    print('Rendering Predicted Mesh')
+                    mesh.show()
+                    print('----------------------')
+
+                if args.f_score:
+                    # Compute F-score
+                    pred_surf, _ = mesh.sample(15000)
+                    f_score = kal.metrics.point.f_score(gt_surf, pred_surf, extend=False)
+                    f_epoch += (f_score / gt_occ.shape[0])
+            i += 1
```
Caenorst, line 142: Are you sure about the position of `i += 1` ? There is a `continue` at line 120 which would skip the increment.
The `i` used at line 123 and 124 would then not be incremented, is this behavior expected ? If not I would suggest using enumerate and avoid the use of `continue`

If it's an error please use

examples/ImageRecon/OccNet/utils.py
```diff
@@ -45,45 +52,47 @@ def get_prior_z(cfg, device, **kwargs):
...
+    elif elem_module == 'numpy':
+        return torch.tensor(np.stack(batch))
+    elif isinstance(elem, container_abcs.Mapping):
+        return {key: collate_fn([d[key] for d in batch]) for key in elem}
+    return batch
 
 
-def extract_mesh(occ_hat, model, c=None, stats_dict=dict()):
-    
+def extract_mesh(occ_hat, model, z, c=None, padding=0.1, threshold=0.2):
```
Caenorst, line 85: `model` is not used in this function

examples/ImageRecon/OccNet/train.py
```diff
@@ -12,234 +12,248 @@
...
+from tqdm import tqdm
 import torch
-from torch.autograd import Variable
-import torch.optim as optim
 from torch.utils.data import DataLoader
-import sys
-from utils import preprocess
-from tqdm import tqdm
 import torch.nn.functional as F
+from torch.utils.tensorboard import SummaryWriter
```
Caenorst, line 23: Can we not use SummaryWriter ? just print logs (or print them in a file). This add a dependency on the example and I don't think it shows any Kaolin specific feature.

kaolin/conversions/voxelgridconversions.py
```diff
@@ -41,7 +41,7 @@
 import torch.nn.functional as F
 import numpy as np
 import kaolin as kal
-import trimesh
+from kaolin import mcubes
```
Caenorst, line 44: Is swapping to this cython version helping for the training ? Because this is changing also a lot of other examples.

PR #198, opened 2020-04-08
add nuscenes as kaolin dataset

jonahthelion:
Add nuscenes as a kaolin dataset. The dataset supports the standard nuscenes train/val split. Samples from the dataset contain a pointcloud in the frame of the ego car and 3D bounding box labels. Users are able to specify how many pointcloud sweeps they would like to accumulate.

### DIFF
--- a/kaolin/datasets/nusc.py
+++ b/kaolin/datasets/nusc.py
@@ -1,0 +1,177 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


Caenorst:
Can you add a little description ?

Also it seems that there is a bunch of dependencies (`pyquaternion`, `nuscenes`), can you add their installation in the Dockerfile ?

Also we need to have a test script for this new dataset, in `tests/datasets/`

kaolin/datasets/nusc.py
```diff
@@ -0,0 +1,177 @@
+# Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
```
Caenorst: It's 2020 now :p

Caenorst (commented):
Just need an update for copyright year and we are good to go :)

Caenorst (approved):
LGTM, thanks ! :)

PR #199, opened 2020-04-10
fix ImageRecon examples shapenet datasets load codes

dkasuga:
fixed shapenet datasets loading codes in order to fit with methods in /kaolin/datasets/shapenet.py


shumash:

PR #201, opened 2020-04-17
silence deprecation warnings

jasongorski:
Deprecation warnings ought to be addressed, however in the interim, silencing them will allow spotting more serious issues at build time before they're committed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@ import os
 import io
 import logging
 from setuptools import setup, find_packages
+import copy
 
 import torch
 from torch.utils.cpp_extension import BuildExtension, CUDAExtension, CppExtension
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -70,6 +71,24 @@ def read(*names, **kwargs):
...
+    if 'extra_compile_args' in kwargs:
+        kwargs['extra_compile_args'] += FLAGS
+    else:
+        kwargs['extra_compile_args'] = FLAGS
+    return CUDAExtension(*args, **kwargs)
+
+
+class KaolinBuildExtension(BuildExtension):
+    def build_extensions(self):
+        FLAG_BLACKLIST = ['-Wstrict-prototypes']
```
Caenorst, line 86: Is this a common flag to be blacklisted ? Otherwise we can just fix the function declarations ?

jasongorski: Filtering -Wstrict-prototypes is to silence these

`cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++`

Caenorst (approved):
LGTM, thanks for the contribution ! :)

PR #203, opened 2020-04-20
Delay import error on nuscenes

Caenorst:
Importing Kaolin would fail if you don't have nuscenes-devkit, even if you are not using NuscDetection
This PR is fixing that by catching the ImportError and raising it when using NuscDetection.


krrish94 (approved):
LGTM, thanks!

PR #205, opened 2020-04-21
Implement MeshCNN

krrish94:
* Implement MeshCNN in Kaolin
* Also port a fix for adjacency-info from @Caenorst 

### DIFF
--- a/kaolin/models/meshcnn.py
+++ b/kaolin/models/meshcnn.py
@@ -1,0 +1,994 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#
+# MIT License
+
+# Copyright (c) 2019 Rana Hanocka
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+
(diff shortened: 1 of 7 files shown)


kaolin/models/meshcnn.py
```diff
@@ -0,0 +1,994 @@
...
+import torch.nn.functional as F
+
+__all__ = [
+    "MeshCNNClassifier",
+    "compute_face_normals_and_areas",
+    "extract_meshcnn_features",
+]
+
+
+def compute_face_normals_for_mesh(mesh):
```
Caenorst, line 54: This function have a very similar name than https://github.com/NVIDIAGameWorks/kaolin/blob/83002fedc67e1c9112a8f834ffb4f8a890e6042a/kaolin/rep/TriangleMesh.py#L154 but a different computation (can you explain ?)

kaolin/models/meshcnn.py
```diff
@@ -0,0 +1,994 @@
...
+
+def compute_face_normals_and_areas(mesh):
+    r"""Compute face normals and areas for an input kaolin.rep.TriangleMesh object.
+    """
+    face_normals = torch.cross(
+        mesh.vertices[mesh.faces[:, 1]] - mesh.vertices[mesh.faces[:, 0]],
+        mesh.vertices[mesh.faces[:, 2]] - mesh.vertices[mesh.faces[:, 1]],
+    )
+    face_normal_lengths = face_normals.norm(p=2, dim=-1)
+    face_normals = face_normals / face_normal_lengths[..., None]
```
Caenorst, line 92: `unqueeze(-1)` at line 61 and `[..., None]` here ?

tests/examples/run/test_classification.sh
```diff
@@ -8,3 +8,8 @@ python ${KAOLIN_HOME}/examples/Classification/pointcloud_classification.py \
     --modelnet-root /data/ModelNet10/ \
     --categories desk dresser \
     --epochs 1 --transforms-device cpu
+
+python ${KAOLIN_HOME}/examples/Classification/mesh_classification.py \
```
Caenorst, line 12: In tests the CI catch the returned value of the script, if you run those two commands like that if the first command fails the return value is gonna be overriden by this one.

I recommend using command1 && command2

Caenorst (changes requested):
Pre-review.

On kaolin/models/meshcnn.py:
We need to have documentation on all the functions (especially for the arguments).
We need to have unit tests for all those functions.
I know is currently not enforced but please avoid using raw string in documentation if you don't need it.

kaolin/models/meshcnn.py
krrish94: As the comment in `compute_face_normals` in `rep/TriangleMesh.py` says, we have a "more robust" normal computation scheme, where we average normals for each vertex to compute the face normal.

Here, I've chosen this computation (as opposed to Kaolin's impl) to exactly align with the feature computation scheme employed in MeshCNN.

krrish94:
Also, for the docstrings, I currently added docstrings only for functions that are exposed externally. All other functions are used only within this file.

Still, I think it's reasonable to add docstrings and unittests for all modules regardless. Will do

Caenorst (approved):
LGTM !

PR #207, opened 2020-04-22
remove nuscenes temporarily

Caenorst:
nuscenes installation have too many dependencies that can break Kaolin installation.
To remove from now until we figure out how to solve that.


Jean-Francois-Lafleche (approved):
LGTM

PR #208, opened 2020-04-23
Update dependencies

Caenorst:
Make Dockerfile, setup.py and README more explicit about the dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,8 @@
+.. image:: https://www.repostatus.org/badges/latest/wip.svg
+  :target: https://www.repostatus.org/#wip
+.. image:: https://img.shields.io/badge/License-Apache%202.0-brightgreen.svg
+  :target: https://github.com/NVIDIAGameWorks/kaolin/blob/master/COPYRIGHT
+
 <p align="center">
     <img src="assets/kaolin.png">
 </p>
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -192,11 +192,13 @@ def get_extensions():
 
 def get_requirements():
     return [
+        'torch>=1.2, <1.5',
```
krrish94: `torch` already gets used before program execution gets here, so it's moot to have this check in `get_requirements()`

krrish94: It might be better off to have a `torch` version check somewhere around the top of this file, and then throw a more descriptive error message?

setup.py
```diff
@@ -205,6 +207,7 @@ def get_requirements():
         'pptk',
```
krrish94, line 212: AFAIK, `pptk` newer versions have been causing viz errors as well. Possibly in a later PR, let's identify a stable version of `pptk` that builds and works w/o hassle.

README.md
```diff
@@ -7,9 +12,11 @@
 
 **[Documentation](https://kaolin.readthedocs.io/en/latest/)** | **[Paper](https://arxiv.org/abs/1911.05063)**
 
+_Kaolin is an active work-in-progress, we are working toward a stable beta release. Please Watch our github repository for releases to be notified. To start trying our code today see instructions below._
```
krrish94: Can we color this in red and make boldface?

krrish94 (changes requested):
Minor fixes requested

setup.py
Caenorst: I remove as dependencies and extended the condition for warning messages, but building Kaolin generate a lot logs, I'm worried that user might not see it easily.

setup.py
Caenorst: Will do it in a subsequent PR

README.md
Caenorst: I changed it for bold, coloring is not really possible in Markdown, if necessary I can make an image.

krrish94:
LGTM, we could decide later if we want to abort the build process instead of logging a warning.

krrish94 (approved):
LGTM, thanks

PR #211, opened 2020-04-24
Simplify Dockerfile and fix torch/torchvision dependencies logs on setup.py

Caenorst:
Remove conda install of unecessary package to limit Dockerfile to a bare minimum.
Invert comparison of versions and using proper package_version and logging.warning instead of string and print in setup.py.


krrish94 (approved):
LGTM

PR #212, opened 2020-04-24
Refactor KaolinDataset and transforms

TommyX12:
- Refactored base `KaolinDataset` to return a `namedtuple`, which allows for unpacking, instead of a dictionary.
- Refactored `kaolin.transforms.transforms` to use a common base class that automatically generates `__repr__`.

### DIFF
--- a/kaolin/transforms/transforms.py
+++ b/kaolin/transforms/transforms.py
@@ -36,6 +36,17 @@ from kaolin.transforms import meshfunc
 from kaolin.transforms import voxelfunc
 
 
+def _get_repr(obj):
+    # TODO: Improve hashing of tensors such that shape matters
+    if isinstance(obj, np.ndarray):
+        return hashlib.sha1(obj).hexdigest()
+
+    if isinstance(obj, torch.Tensor):
+        return hashlib.sha1(obj.cpu().numpy()).hexdigest()
+
+    return repr(obj)
+
+
 class Transform(object):
     """Base class for all Kaolin transforms.
 
@@ -61,14 +72,14 @@ class Transform(object):
     """
 
     def __repr__(self):
-        ignored = set(getattr(self.__class__, '__ignored_params__', []))
+        ignored = set(getattr(self, '__ignored_params__', []))
         names = [
             k for k in self.__dict__.keys()
             if k not in ignored
         ]
         names.sort()
         params = ', '.join([
-            '{}={}'.format(name, repr(self.__dict__[name]))
+            '{}={}'.format(name, _get_repr(self.__dict__[name]))
             for name in names
         ])
         return '{}({})'.format(self.__class__.__name__, params)
@@ -204,8 +215,6 @@ class ScalePointCloud(Transform):
 
     """
 
-    __ignored_params__ = ['inplace']
-
     def __init__(self, scf: Union[int, float, torch.Tensor],
                  inplace: Optional[bool] = True):
         self.scf = scf
@@ -236,8 +245,6 @@ class RotatePointCloud(Transform):
 
     """
 
-    __ignored_params__ = ['inplace']
-
     def __init__(self, rotmat: torch.Tensor, inplace: Optional[bool] = True):
         self.rotmat = rotmat
         self.inplace = inplace
@@ -264,12 +271,7 @@ class RealignPointCloud(Transform):
 
     TODO: Example.
 
-    TODO: This transform's repr is incorrect:
-    changing tgt will not cause caches to recompute.
-
     """
-
-    __ignored_params__ = ['tgt', 'inplace']
 
     def __init__(self, tgt: Union[torch.Tensor, PointCloud],
                  inplace: Optional[bool] = True):
@@ -299,8 +301,6 @@ class NormalizePointCloud(Transform):
     TODO: Example.
 
     """
-
-    __ignored_params__ = ['inplace']
 
     def __init__(self, inplace: Optional[bool] = True):
         self.inplace = inplace
@@ -332,8 +332,6 @@ class DownsampleVoxelGrid(Transform):
     TODO: Example.
 
     """
-
-    __ignored_params__ = ['inplace']
 
     def __init__(self, scale: List[int], inplace=True):
         self.scale = scale
@@ -391,8 +389,6 @@ class ThresholdVoxelGrid(Transform):
         inplace (bool, optional): Bool to make the operation in-place.
 
     """
-
-    __ignored_params__ = ['inplace']
 
     def __init__(self, thresh: float, inplace: Optional[bool] = True):
         self.thresh = thresh
@@ -499,8 +495,6 @@ class SampleTriangleMesh(Transform):
                      for small surface areas.
     """
 
-    __ignored_params__ = ['eps']
-
     def __init__(self, num_samples: int, eps: Optional[float] = 1e-10):
         self.num_samples = num_samples
         self.eps = eps
@@ -531,8 +525,6 @@ class NormalizeMesh(Transform):
     TODO: Example.
 
     """
-
-    __ignored_params__ = ['inplace']
 
     def __init__(self, inplace: Optional[bool] = True):
         self.inplace = inplace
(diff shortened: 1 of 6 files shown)


Caenorst (commented):
Can you rebase and run on internal CI ? I'm guessing some of the examples might fail to run.

LGTM otherwise

kaolin/transforms/transforms.py
```diff
@@ -36,7 +36,45 @@
...
-class Compose(object):
+class Transform(object):
+    """Base class for all Kaolin transforms.
+
+    This class generates __repr__ string automatically. Given that all attributes
+    have valid __repr__, it generates a string with the following format:
+    .. code-block::
+       MyClass(a=1, b=2, ...)
+
+    The attributes are sorted alphabetically to ensure determinism.
```
Caenorst, line 58: I don't see where you are ordering alphabetically, I believe dict keys ordering depend on python version.

kaolin/transforms/transforms.py
```diff
@@ -36,7 +36,45 @@
...
+    NOTE:
+    - Since the __repr__ generation depends on the __repr__ of attributes,
+      objects that have incorrect (such as arbitrary objects that output
+      their memory address by default) or indeterministic (such as
+      dictionaries) __repr__ should not be assigned as attribute.
+      If unavoidable, override __repr__ manually.
+    """
+
+    def __repr__(self):
+        ignored = set(getattr(self.__class__, '__ignored_params__', []))
```
Caenorst: Why not just `ignored = set(getattr(self, '__ignored_params__', []))` ? It would works for both class and instance attribute (if for instance there is some conditional params to ignore or something like that).

kaolin/transforms/transforms.py
```diff
@@ -237,8 +264,13 @@ class RealignPointCloud(object):
 
     TODO: Example.
 
+    TODO: This transform's repr is incorrect:
```
Caenorst: I think this TODO is resolved ?

Caenorst (commented):
Should we really ignore the params `eps` and `inplace` ? What if `inplace` is incorrectly used (like it actually modify a tensor that was supposed to be reused), what if `eps` have a strong impact on the output ?

I'm actually struggling to see cases where you can ignore an attribute, if you change an attribute to a transform I assume you actually want to change the output right ? If your attribute for instance is only speed or memory related, I would argue that the most efficient way to have the output is.... to use the cache (so don't change the attributes) !

Do you have an example ?

kaolin/transforms/transforms.py
TommyX12: I believe Python (starting from 3.7) only guarantees dict key to be sorted by insertion order; so I am not sure if that guarantees two objects with the same attributes will always get the same repr. Sorting alphabetically guarantees that :P

kaolin/transforms/transforms.py
TommyX12: That's a good point. My concern is that if the object itself can have `__ignored_params__` then you would also have to put `__ignored_params__` either in the class's `__ignored_params__` or the object's `__ignored_params__`, which might be a little error-prone. Maybe we should avoid this entirely by just using the class attribute?

kaolin/transforms/transforms.py
TommyX12: `tgt` is ignored in the `repr`, but it shouldn't be because it does change the output of the transformation. However I have yet to figure out a way to actually put it into the repr, since it could be a giant tensor.

TommyX12:
> Should we really ignore the params `eps` and `inplace` ? What if `inplace` is incorrectly used (like it actually modify a tensor that was supposed to be reused), what if `eps` have a strong impact on the output ?
> 
> I'm actually struggling to see cases where you can ignore an attribute, if you change an attribute to a transform I assume you actually want to change the output right ? If your attribute for instance is only speed or memory related, I would argue that the most efficient way to have the output is.... to use the cache (so don't change the attributes) !
> 
> Do you have an example ?

That's actually a good point. I will remove `inplace` and `eps` from the list.

kaolin/transforms/transforms.py
Caenorst: Yes, but, where are you enforcing alphabetical order ?

Caenorst: Yes, but, where are you enforcing alphabetical order ?

kaolin/transforms/transforms.py
Caenorst: It's hashed, only once per dataset constructor: https://github.com/NVIDIAGameWorks/kaolin/blob/master/kaolin/datasets/base.py#L90 I think you can just serialize it, you can also hash it.

Caenorst:
> > Should we really ignore the params `eps` and `inplace` ? What if `inplace` is incorrectly used (like it actually modify a tensor that was supposed to be reused), what if `eps` have a strong impact on the output ?
> > I'm actually struggling to see cases where you can ignore an attribute, if you change an attribute to a transform I assume you actually want to change the output right ? If your attribute for instance is only speed or memory related, I would argue that the most efficient way to have the output is.... to use the cache (so don't change the attributes) !
> > Do you have an example ?
> 
> That's actually a good point. I will remove `inplace` and `eps` from the list.

But then is there any use at `__ignore_params__` ?

kaolin/transforms/transforms.py
TommyX12: https://github.com/NVIDIAGameWorks/kaolin/blob/8937233ba30059eb3d003d1add0fad1ee54c4c1a/kaolin/transforms/transforms.py#L69

Caenorst: Missed that line :)

kaolin/transforms/transforms.py
Caenorst: Well it's python, stubborn users will always have possibility to stupid things :D
I would assume that the user who would create his own class would know what he's doing.

The base class is not enforcing class or instantiate attribute neither, so it's not misleading to let the user do whatever he wants

TommyX12: > > > Should we really ignore the params `eps` and `inplace` ? What if `inplace` is incorrectly used (like it actually modify a tensor that was supposed to be reused), what if `eps` have a strong impact on the output ?
> > > I'm actually struggling to see cases where you can ignore an attribute, if you change an attribute to a transform I assume you actually want to change the output right ? If your attribute for instance is only speed or memory related, I would argue that the most efficient way to have the output is.... to use the cache (so don't change the attributes) !
> > > Do you have an example ?
> > 
> > 
> > That's actually a good point. I will remove `inplace` and `eps` from the list.
> 
> But then is there any use at `__ignore_params__` ?

Maybe here?
https://github.com/NVIDIAGameWorks/kaolin/blob/8937233ba30059eb3d003d1add0fad1ee54c4c1a/kaolin/transforms/transforms.py#L108

TommyX12:
> > > Should we really ignore the params `eps` and `inplace` ? What if `inplace` is incorrectly used (like it actually modify a tensor that was supposed to be reused), what if `eps` have a strong impact on the output ?
> > > I'm actually struggling to see cases where you can ignore an attribute, if you change an attribute to a transform I assume you actually want to change the output right ? If your attribute for instance is only speed or memory related, I would argue that the most efficient way to have the output is.... to use the cache (so don't change the attributes) !
> > > Do you have an example ?
> > 
> > 
> > That's actually a good point. I will remove `inplace` and `eps` from the list.
> 
> But then is there any use at `__ignore_params__` ?

Maybe here?
https://github.com/NVIDIAGameWorks/kaolin/blob/8937233ba30059eb3d003d1add0fad1ee54c4c1a/kaolin/transforms/transforms.py#L108

Caenorst:
> Maybe here?
> https://github.com/NVIDIAGameWorks/kaolin/blob/8937233ba30059eb3d003d1add0fad1ee54c4c1a/kaolin/transforms/transforms.py#L108

Fair, although I would question why somebody would suddenly change cache_dir ^^'

kaolin/transforms/transforms.py
TommyX12: Good point. Fixed.

kaolin/transforms/transforms.py
TommyX12: Done.

Caenorst (approved):
LGTM, great feature ! :)

PR #213, opened 2020-04-27
Fix README badges

krrish94:
Badges in `README` are incorrectely rendered (Markdown syntax errors)


Caenorst (approved):
LGTM, thanks for the fix

PR #214, opened 2020-04-27
Add a test to check if classification example is running and fix

Caenorst:
Fix examples/pointcloud_classification.py
new features to the example:
   1) --viz-test argument to avoid plot attempt in CI
   2) Speedup using dataloader multiprocess (speedup variable, x4 on my workstation)
   2) replace --device to be --transforms-device, because depending on your processor applying transforms on GPU may still be more efficient.

Comments:
1) examples/Classification/pointcloud_classification_engine.py is not fixed yet
2) the test will probably need some renaming after #205 

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,8 +13,8 @@ test:
     - kaolin
   stage: test
   script:
+    - flake8 --config=.flake8 .
     - source setenv.sh
     - python setup.py develop
+    - pytest --cov=kaolin/ tests/
     - ./tests/examples/run/test_classification.sh
-      #- flake8 --config=.flake8 .
-      #- pytest --cov=kaolin/ tests/
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,7 +13,8 @@ test:
     - kaolin
   stage: test
   script:
-    - flake8 --config=.flake8 .
     - source setenv.sh
     - python setup.py develop
-    - pytest --cov=kaolin/ tests/
+    - ./tests/examples/run/test_classification.sh
+      #- flake8 --config=.flake8 .
+      #- pytest --cov=kaolin/ tests/
--- a/examples/Classification/pointcloud_classification.py
+++ b/examples/Classification/pointcloud_classification.py
@@ -66,7 +66,6 @@ for e in range(args.epochs):
 
     model.train()
 
-    optimizer.zero_grad()
     for idx, batch in enumerate(tqdm(train_loader)):
         category = batch['attributes']['category'].cuda()
         pred = model(batch['data'].cuda())
@@ -74,6 +73,7 @@ for e in range(args.epochs):
         train_loss += loss.item()
         loss.backward()
         optimizer.step()
+        optimizer.zero_grad()
 
         # Compute accuracy
         pred_label = torch.argmax(pred, dim=1)
--- a/examples/Classification/pointcloud_classification.py
+++ b/examples/Classification/pointcloud_classification.py
@@ -1,13 +1,14 @@
 import argparse
+import time
 
 from tqdm import tqdm
 import torch
 from torch.utils.data import DataLoader
+from torch.multiprocessing import cpu_count
 
 from kaolin.datasets import ModelNet
 from kaolin.models.PointNet import PointNetClassifier
 import kaolin.transforms as tfs
-from utils import visualize_batch
 
 parser = argparse.ArgumentParser()
 parser.add_argument('--modelnet-root', type=str, help='Root directory of the ModelNet dataset.')
@@ -16,28 +17,43 @@ parser.add_argument('--num-points', type=int, default=1024, help='Number of poin
 parser.add_argument('--epochs', type=int, default=10, help='Number of train epochs.')
 parser.add_argument('-lr', '--learning-rate', type=float, default=1e-3, help='Learning rate.')
 parser.add_argument('--batch-size', type=int, default=12, help='Batch size.')
-parser.add_argument('--device', type=str, default='cuda', help='Device to use.')
+parser.add_argument('--viz-test', action='store_true', help='Visualize an output of a test sample')
+parser.add_argument('--transforms-device', type=str, default='cuda', help='Device to use.')
 
 args = parser.parse_args()
 
 
+def to_device(inp):
+    inp.to(args.transforms_device)
+    return inp
+
 transform = tfs.Compose([
+    to_device,
     tfs.TriangleMeshToPointCloud(num_samples=args.num_points),
     tfs.NormalizePointCloud()
 ])
 
+if args.transforms_device == 'cuda':
+    num_workers = 0
+    pin_memory = False
+else:
+    num_workers = cpu_count()
+    pin_memory = True
+
 train_loader = DataLoader(ModelNet(args.modelnet_root, categories=args.categories,
-                                   split='train', transform=transform, device=args.device),
-                          batch_size=args.batch_size, shuffle=True)
+                                   split='train', transform=transform),
+                          batch_size=args.batch_size, shuffle=True,
+                          num_workers=num_workers, pin_memory=pin_memory)
 
 val_loader = DataLoader(ModelNet(args.modelnet_root, categories=args.categories,
-                                 split='test', transform=transform, device=args.device),
-                        batch_size=args.batch_size)
+                                 split='test', transform=transform),
+                        batch_size=args.batch_size,
+                        num_workers=num_workers, pin_memory=pin_memory)
 
-model = PointNetClassifier(num_classes=len(args.categories)).to(args.device)
+model = PointNetClassifier(num_classes=len(args.categories)).to('cuda')
 optimizer = torch.optim.Adam(model.parameters(), lr=args.learning_rate)
 criterion = torch.nn.CrossEntropyLoss()
-
+start_time = time.time()
 for e in range(args.epochs):
 
     print('###################')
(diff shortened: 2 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -13,7 +13,8 @@ test:
     - kaolin
   stage: test
   script:
-    - flake8 --config=.flake8 .
     - source setenv.sh
     - python setup.py develop
-    - pytest --cov=kaolin/ tests/
+    - ./tests/examples/run/test_classification.sh
+      #- flake8 --config=.flake8 .
+      #- pytest --cov=kaolin/ tests/
```
krrish94: Can delete

examples/Classification/pointcloud_classification.py
```diff
@@ -16,28 +17,43 @@
...
+    to_device,
     tfs.TriangleMeshToPointCloud(num_samples=args.num_points),
     tfs.NormalizePointCloud()
 ])
 
+if args.transforms_device == 'cuda':
+    num_workers = 0
+    pin_memory = False
+else:
+    num_workers = cpu_count()
```
krrish94: Might not be a good idea to use all cpus? Like in pytorch examples, let's set the default to a nominal value (like `1` or `2`) and let the user decide on the number of threads to spawn?

cf. https://github.com/pytorch/examples/blob/master/mnist/main.py#L100

examples/Classification/pointcloud_classification.py
```diff
@@ -16,28 +17,43 @@
...
 
 val_loader = DataLoader(ModelNet(args.modelnet_root, categories=args.categories,
-                                 split='test', transform=transform, device=args.device),
-                        batch_size=args.batch_size)
+                                 split='test', transform=transform),
+                        batch_size=args.batch_size,
+                        num_workers=num_workers, pin_memory=pin_memory)
 
-model = PointNetClassifier(num_classes=len(args.categories)).to(args.device)
+model = PointNetClassifier(num_classes=len(args.categories)).to('cuda')
```
krrish94, line 53: We might want to run `cpu` tests too. How about adding a `device` commandline arg that specifies the device to perform training on? (Not high priority for the current PR)

examples/Classification/pointcloud_classification.py
```diff
@@ -16,28 +17,43 @@
...
 optimizer = torch.optim.Adam(model.parameters(), lr=args.learning_rate)
 criterion = torch.nn.CrossEntropyLoss()
-
+start_time = time.time()
```
krrish94, line 56: How about using `tqdm` / `trange`? We already have it as a dependency.

Downside is we might not want to litter logs with progress bars. I'm fine with either way.

examples/Classification/pointcloud_classification.py
```diff
@@ -52,15 +68,16 @@
 
     optimizer.zero_grad()
```
krrish94: I'd prefer moving this to below `optimizer.step()`. That in general helps keep code modular (i.e., all backward/optimizer related code is at one place).

examples/Classification/pointcloud_classification.py
```diff
@@ -74,24 +91,28 @@
...
+            loss = criterion(pred, category.view(-1))
             val_loss += loss.item()
 
             # Compute accuracy
             pred_label = torch.argmax(pred, dim=1)
-            val_accuracy += torch.mean((pred_label == batch[1].view(-1)).float()).cpu().item()
+            val_accuracy += torch.mean((pred_label == category.view(-1)).float()).cpu().item()
             num_batches += 1
 
     print('Val loss:', val_loss / num_batches)
```
krrish94, line 104: Add tiny `eps` to the denominator? (To prevent unforeseen errors, eg. if val batchsize is `0` by accident/intention)

examples/Classification/pointcloud_classification.py
```diff
@@ -74,24 +91,28 @@
...
     print('Val accuracy:', val_accuracy / num_batches)
```
krrish94, line 105: Same as above comment (tiny `eps` to denominator)

examples/Classification/pointcloud_classification.py
```diff
@@ -74,24 +91,28 @@
...
 test_loader = DataLoader(ModelNet(args.modelnet_root, categories=args.categories,
-                                  split='test', transform=transform, device=args.device),
-                         shuffle=True, batch_size=15)
+                                  split='test', transform=transform),
+                         shuffle=True, batch_size=15, num_workers=num_workers, pin_memory=pin_memory)
 
-test_batch, labels = next(iter(test_loader))
-preds = model(test_batch)
+test_batch = next(iter(test_loader))
+preds = model(test_batch['data'].cuda())
```
krrish94, line 113: `.to(device)` instead of `cuda()` (only if following an above comment asking for multi-device tests)

kaolin/transforms/meshfunc.py
```diff
@@ -65,8 +65,8 @@ def sample_triangle_mesh(vertices: torch.Tensor, faces: torch.Tensor,
     # We want the last dimension of vertices to be of shape 3.
     helpers._assert_shape_eq(vertices, (-1, 3), dim=-1)
 
-    dist_uni = torch.distributions.Uniform(torch.tensor([0.]).to(
-        vertices.device), torch.tensor([1.]).to(vertices.device))
+    dist_uni = torch.distributions.Uniform(torch.zeros((1,), device=vertices.device),
```
krrish94, line 68: Unrelated to PR, so suggest removing. What do you feel?

krrish94 (changes requested):
Thanks, a few minor fixes and we should be good.

.gitlab-ci.yml
Caenorst: I uncommented and moved it back where it were

krrish94: Ah yes, my bad :)

examples/Classification/pointcloud_classification.py
Caenorst: I disagree, this is not an unit test but an example, this is to show the standard usage of Kaolin, which is with GPU, I wouldn't imagine somebody starting to do DL for 3D with CPU only.

On top of that it will add big work on our CI because we will have to run that as well.

krrish94: Agree on CI load; I have a few reservations about the GPU-only advocacy, but it's cool if we currently do GPU-only.

examples/Classification/pointcloud_classification.py
krrish94: Resolving this, as we aren't doing CPU tests now

examples/Classification/pointcloud_classification.py
Caenorst: Later we'd like to grep those metrics (accuracy, training time) `tests/examples/*.sh` to check that we don't have speed regression. Not sure I can do that easily with tqdm logs.

krrish94: Agree; tqdm litters a lot

examples/Classification/pointcloud_classification.py
Caenorst: How can batchsize be 0 other than if the user purposely set it to be 0 ? If he does Dataloader would fail to construct `ValueError: batch_size should be a positive integer value, but got batch_size=0`

Caenorst: I assume that you mean `num_batches`, not `batchsize`. How can that happens in a way that requires us to support it ?

examples/Classification/pointcloud_classification.py
Caenorst: I'm not saying that Kaolin is GPU-only, I'm just saying that we should use CPU where it's pertinent, I may be biased but I would not not advise any user to do 3D DL with CPU-only.

Unit tests for ops should still be executed with CPU if supported.

kaolin/transforms/meshfunc.py
Caenorst: It's (more or less) part of the speedup / clean, it avoid one unecessary transfer to GPU.

examples/Classification/pointcloud_classification.py
Caenorst: done

examples/Classification/pointcloud_classification.py
Caenorst: Done

This strongly depend on the training loop bottleneck: 
https://github.com/pytorch/examples/blob/master/dcgan/main.py#L19
https://github.com/pytorch/examples/blob/master/imagenet/main.py#L33
https://github.com/pytorch/examples/blob/master/super_resolution/main.py#L20

Our conversion function right now is just not efficient enough on CPU to be interesting with such a low number of workers, we will be working on that later

examples/Classification/pointcloud_classification.py
krrish94: Doesn't usually happen. I can see two cases in which this happens.
1. Missing data for particular category (could happen on CI, esp if we move to multiple machines)
2. User wilfully deletes data for particular category, to skip validation, as a hack

Again, I don't think this is super-important, as this isn't the intended use-case. I'll go by whatever you recommend.

krrish94 (approved):
LGTM, thanks for the clarifications on my comments

examples/Classification/pointcloud_classification.py
Caenorst: 1) We definitely want it to crash ! Otherwise the missing data will go unnoticed.
2) We don't recommend people "hacking" our examples. (Neither training without cross-validation).

PR #236, opened 2020-04-30
Establish softras parity with original impl

krrish94:
* Ensure softras impl exactly matches original code release
* Add regression test to compare with rendered outputs from original impl
* Add examples and README


examples/renderers/README.md
```diff
@@ -39,12 +39,65 @@ DIB-Renderer
 }
 ```
 
-## Building
+
```
Caenorst: Shouldn't this be in `examples/renderers/softras/README.md` ?

examples/renderers/softras_simple_render.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # Initialize the soft rasterizer.
+    renderer = kaolin.graphics.SoftRenderer(camera_mode="look_at", device="cuda:0")
+
+    # Camera settings.
+    camera_distance = 2.  # Distance of the camera from the origin (i.e., center of the object).
+    elevation = 30.       # Angle of elevation
+
+    # Infer the base path of the kaolin repo
+    KAOLIN_ROOT = os.path.join(os.path.dirname(__file__), "..", "..")
```
Caenorst: Instead of going to the root, just copy paste the banana obj and do everything in the same examples/renderers/SoftRas folder

examples/renderers/softras_simple_render.py
```diff
@@ -0,0 +1,107 @@
...
+    # Read in the input mesh.
+    infile = os.path.join(KAOLIN_ROOT, "tests", "graphics", "banana.obj")
+    mesh = kaolin.rep.TriangleMesh.from_obj(infile)
+
+    # Output filename (to write out a rendered .gif to).
+    outfile = os.path.join(KAOLIN_ROOT, "examples", "renderers", "softras_render.gif")
+
+    # Extract the vertices, faces, and texture the mesh (currently color with white).
+    vertices = mesh.vertices.float()
+    faces = mesh.faces.long()
```
Caenorst: Why ? you just loaded it using https://github.com/krrish94/kaolin-bleed/blob/softras-parity/kaolin/rep/Mesh.py#L164-L165 , so we already know mesh.vertices are `torch.float` and faces are `torch.long`

examples/renderers/softras_simple_render.py
```diff
@@ -0,0 +1,107 @@
...
+    face_textures = faces.clone()
```
Caenorst: face_textures is unused

examples/renderers/softras_simple_render.py
```diff
@@ -0,0 +1,107 @@
...
+    vertices = vertices[None, :, :].cuda()
+    faces = faces[None, :, :].cuda()
+    face_textures = face_textures[None, :, :].cuda()
```
Caenorst: face_textures is unused

examples/renderers/softras_texture_optimization.py
```diff
@@ -0,0 +1,142 @@
...
+    # Extract the vertices, faces, and texture the mesh (currently color with white).
+    vertices = mesh.vertices.float()
+    faces = mesh.faces.long()
+    face_textures = faces.clone()
+    vertices = vertices[None, :, :].cuda()
+    faces = faces[None, :, :].cuda()
+    face_textures = face_textures[None, :, :].cuda()
+    textures = torch.ones(1, faces.shape[1], 2, 3, dtype=torch.float32, device="cuda:0")
+
+    # Normalize vertices
```
Caenorst: I believe this just "centering" and not "normalizing"
(TODO: centering / normalizing would be nice ops in the core)

examples/renderers/softras_simple_render.py
```diff
@@ -0,0 +1,107 @@
...
+    vertices_min = vertices.min()
+    vertices_middle = (vertices_max + vertices_min) / 2.
+    vertices = vertices - vertices_middle
+    # Scale the vertices slightly (so that they occupy a sizeable image area).
+    # Skip if using models other than the banana.obj file.
+    coef = 5
+    vertices = vertices * coef
+
+    # Loop over a set of azimuth angles, and render the image.
+    loop = loop = tqdm.tqdm(list(range(0, 360, 6)))
```
Caenorst: I don't understand this syntax, can you explain me ?
You can directly use `tqdm(range(...))`

examples/renderers/softras_simple_render.py
```diff
@@ -0,0 +1,107 @@
...
+    textures = torch.cat(
+        (
+            torch.ones(1, faces.shape[1], 2, 1, dtype=torch.float32, device="cuda:0"),
+            torch.ones(1, faces.shape[1], 2, 1, dtype=torch.float32, device="cuda:0"),
+            torch.zeros(1, faces.shape[1], 2, 1, dtype=torch.float32, device="cuda:0"),
+        ),
+        dim=-1,
+    )
+
+    # Normalize vertices
```
Caenorst: I believe this just "centering" and not "normalizing"
(TODO: centering / normalizing would be nice ops in the core)

examples/renderers/softras_texture_optimization.py
```diff
@@ -0,0 +1,142 @@
...
+    # Initialize the soft rasterizer.
+    renderer = kaolin.graphics.SoftRenderer(camera_mode="look_at", device="cuda:0")
+
+    # Camera settings.
+    camera_distance = 2.  # Distance of the camera from the origin (i.e., center of the object)
+    elevation = 30.       # Angle of elevation
+    azimuth = 0.          # Azimuth angle
+
+    # Infer the base path of the kaolin repo
+    KAOLIN_ROOT = os.path.join(os.path.dirname(__file__), "..", "..")
```
Caenorst: Instead of going to the root, just copy paste the banana obj and do everything in the same examples/renderers/SoftRas folder.

examples/renderers/softras_texture_optimization.py
```diff
@@ -0,0 +1,142 @@
...
+    mesh = kaolin.rep.TriangleMesh.from_obj(os.path.join(KAOLIN_ROOT, "tests", "graphics", "banana.obj"))
+
+    # Output filename to write out a rendered .gif to, showing the progress of optimization.
+    progressfile = os.path.join(KAOLIN_ROOT, "examples", "renderers", "texture_optimization_progress.gif")
+    # Output filename to write out a rendered .gif file to, rendering the optimized mesh.
+    outfile = os.path.join(KAOLIN_ROOT, "examples", "renderers", "texture_optimization_output.gif")
+
+    # Extract the vertices, faces, and texture the mesh (currently color with white).
+    vertices = mesh.vertices.float()
+    faces = mesh.faces.long()
```
Caenorst: Why ? you just loaded it using https://github.com/krrish94/kaolin-bleed/blob/softras-parity/kaolin/rep/Mesh.py#L164-L165 , so we already know mesh.vertices are `torch.float` and faces are `torch.long`

examples/renderers/softras_texture_optimization.py
```diff
@@ -0,0 +1,142 @@
...
+    face_textures = faces.clone()
```
Caenorst: `face_textures` is unused

examples/renderers/softras_texture_optimization.py
```diff
@@ -0,0 +1,142 @@
...
+    vertices = vertices[None, :, :].cuda()
+    faces = faces[None, :, :].cuda()
+    face_textures = face_textures[None, :, :].cuda()
```
Caenorst: `face_textures` is unused

examples/renderers/softras_texture_optimization.py
```diff
@@ -0,0 +1,142 @@
...
+        if i % 5 == 0:
+            # TODO: Add functionality to write to gif output file.
+            tqdm.write("Loss: " + str(loss.item()) + "\n")
+            img = rgba[0].permute(1, 2, 0).detach().cpu().numpy()
+            writer.append_data((255 * img).astype(np.uint8))
+    writer.close()
+
+    # Write optimized mesh to output file.
+    writer = imageio.get_writer(outfile, mode="I")
+    for azimuth in tqdm(list(range(0, 360, 6))):
```
Caenorst: You can directly use tqdm(range(...))

examples/renderers/softras_vertex_optimization.py
```diff
@@ -0,0 +1,144 @@
...
+    # Initialize the soft rasterizer.
+    renderer = kaolin.graphics.SoftRenderer(camera_mode="look_at", device="cuda:0")
+
+    # Camera settings.
+    camera_distance = 2.  # Distance of the camera from the origin (i.e., center of the object)
+    elevation = 30.       # Angle of elevation
+    azimuth = 0.          # Azimuth angle
+
+    # Infer the base path of the kaolin repo
+    KAOLIN_ROOT = os.path.join(os.path.dirname(__file__), "..", "..")
```
Caenorst: Instead of going to the root, just copy paste the banana obj and do everything in the same examples/renderers/SoftRas folder.

examples/renderers/softras_vertex_optimization.py
```diff
@@ -0,0 +1,144 @@
...
+    mesh = kaolin.rep.TriangleMesh.from_obj(os.path.join(KAOLIN_ROOT, "tests", "model.obj"))
+
+    # Output filename to write out a rendered .gif to, showing the progress of optimization.
+    progressfile = os.path.join(KAOLIN_ROOT, "examples", "renderers", "vertex_optimization_progress.gif")
+    # Output filename to write out a rendered .gif file to, rendering the optimized mesh.
+    outfile = os.path.join(KAOLIN_ROOT, "examples", "renderers", "vertex_optimization_output.gif")
+
+    # Extract the vertices, faces, and texture the mesh (currently color with white).
+    vertices = mesh.vertices.float()
+    faces = mesh.faces.long()
```
Caenorst: Why ? you just loaded it using https://github.com/krrish94/kaolin-bleed/blob/softras-parity/kaolin/rep/Mesh.py#L164-L165 , so we already know mesh.vertices are torch.float and faces are torch.long

examples/renderers/softras_vertex_optimization.py
```diff
@@ -0,0 +1,144 @@
...
+    face_textures = faces.clone()
```
Caenorst: `face_textures` is unused

examples/renderers/softras_vertex_optimization.py
```diff
@@ -0,0 +1,144 @@
...
+    vertices = vertices[None, :, :].cuda()
+    faces = faces[None, :, :].cuda()
+    face_textures = face_textures[None, :, :].cuda()
```
Caenorst: `face_textures` is unused

examples/renderers/softras_vertex_optimization.py
```diff
@@ -0,0 +1,144 @@
...
+    textures = torch.cat(
+        (
+            torch.ones(1, faces.shape[1], 2, 1, dtype=torch.float32, device="cuda:0"),
+            torch.ones(1, faces.shape[1], 2, 1, dtype=torch.float32, device="cuda:0"),
+            torch.zeros(1, faces.shape[1], 2, 1, dtype=torch.float32, device="cuda:0"),
+        ),
+        dim=-1,
+    )
+
+    # # TODO: Normalize vertices (Use kaolin functionality to do this)
```
Caenorst: Looks like I'm not the only one who think that it should be a Kaolin functionality :)

examples/renderers/softras_vertex_optimization.py
```diff
@@ -0,0 +1,144 @@
...
+        if i % 20 == 0:
+            # TODO: Add functionality to write to gif output file.
+            tqdm.write(f"Loss: {loss.item():.5}")
+            img = rgba[0].permute(1, 2, 0).detach().cpu().numpy()
+            writer.append_data((255 * img).astype(np.uint8))
+    writer.close()
+
+    # Write optimized mesh to output file.
+    writer = imageio.get_writer(outfile, mode="I")
+    for azimuth in tqdm(list(range(0, 360, 6))):
```
Caenorst: You can directly use `tqdm(range(...))`

kaolin/graphics/Lighting.py
```diff
@@ -255,12 +264,10 @@ def compute_directional_light(
     if directional_intensity == 0:
         return light
 
-    # Unwrap the batchsize dimension (for easier vectorization).
-    faces = face_vertices.reshape(batchsize * num_faces, 3, 3)
     # Compute face normals.
```
Caenorst, line 267: https://github.com/NVIDIAGameWorks/kaolin/blob/master/kaolin/rep/TriangleMesh.py#L154
https://github.com/NVIDIAGameWorks/kaolin/blob/master/kaolin/models/meshcnn.py#L54

I don't understand why we won't use one of those

kaolin/graphics/SoftRenderer.py
```diff
@@ -664,7 +640,37 @@ def textures_to_faces(self, textures, faces):
...
+                            elevation: Union[int, float, torch.Tensor],
+                            azimuth: Union[int, float, torch.Tensor],
+                            degrees: bool = True):
+        r"""Compute the position of the camera given azimuth and elevation angles,
+        and the distance of the camera from the origin (i.e., spherical coordinates).
+
+        Args:
+            distance (int, float, or torch.Tensor): Distance of the camera from the origin.
+            azimuth (int, float, or torch.Tensor): Azimuth angle.
+            elevation (int, float, or torch.Tensor): Elevation angle.
```
Caenorst: I don't think supporting `int` makes sense

tests/graphics/test_softras.py
```diff
@@ -16,19 +16,34 @@
...
+    filename_input = "tests/graphics/banana.obj"
+
     mesh = kaolin.rep.TriangleMesh.from_obj(filename_input)
-    vertices = mesh.vertices.float()
+
+    vertices = mesh.vertices
     faces = mesh.faces.long()
-    face_textures = faces.clone()
-    vertices = vertices[None, :, :].cuda()
+    face_textures = (faces).clone()
```
Caenorst: `face_textures` is unused. (and why adding `()` ?)

tests/graphics/test_softras.py
```diff
@@ -16,19 +16,34 @@
...
-    filename_input = 'tests/model.obj'
-    
+    renderer = kaolin.graphics.SoftRenderer(camera_mode="look_at", device=device)
+    filename_input = "tests/graphics/banana.obj"
+
     mesh = kaolin.rep.TriangleMesh.from_obj(filename_input)
-    vertices = mesh.vertices.float()
+
+    vertices = mesh.vertices
     faces = mesh.faces.long()
```
Caenorst: `face` is already `long`

tests/graphics/test_softras.py
```diff
@@ -16,19 +16,34 @@
...
-    face_textures = faces.clone()
-    vertices = vertices[None, :, :].cuda()
+    face_textures = (faces).clone()
+    vertices = vertices[None, :, :].cuda()  
     faces = faces[None, :, :].cuda()
-    face_textures = face_textures[None, :, :].cuda()
+    face_textures[None, :, :].cuda()
```
Caenorst: `face_textures` is unused.

Caenorst (changes requested):
I think all of that should be inside `examples/renderers/softras/` get ride of previous examples as they are irrelevant.

examples/renderers/softras_simple_render.py
krrish94: Fair enough. Also, how about creating an `sampledata` folder inside `examples/renderers`. We could then share the same `banana.obj` file across DIB-R and NMR examples

examples/renderers/softras_simple_render.py
krrish94: Good catch, this was copy-pasted from the old softras example

examples/renderers/softras_simple_render.py
krrish94: Good catch, this was copy-pasted from the old softras example

examples/renderers/softras_simple_render.py
krrish94: Good catch, this was copy-pasted from the old softras example

examples/renderers/softras_texture_optimization.py
krrish94: Yeah, makes sense to support as transformations

examples/renderers/softras_simple_render.py
krrish94: This was copy-pasted from the old softras example; will change

examples/renderers/softras_simple_render.py
krrish94: Makes sense to support (we already have a transform that's quite close)

examples/renderers/softras_texture_optimization.py
krrish94: How about placing it inside a `sampledata` folder inside `examples/renderers`? It can then be used across DIB-R, NMR, etc

examples/renderers/softras_texture_optimization.py
krrish94: Good catch, this was copy-pasted from the old softras example

examples/renderers/softras_texture_optimization.py
krrish94: Good catch, this was copy-pasted from the old softras example

examples/renderers/softras_texture_optimization.py
krrish94: Good catch, this was copy-pasted from the old softras example

examples/renderers/softras_texture_optimization.py
krrish94: this was copy-pasted from the old softras example, will change

examples/renderers/softras_vertex_optimization.py
krrish94: How about placing it inside a sampledata folder inside examples/renderers? It can then be used across DIB-R, NMR, etc

examples/renderers/softras_vertex_optimization.py
krrish94: Good catch, this was copy-pasted from the old softras example

examples/renderers/softras_vertex_optimization.py
krrish94: Good catch, this was copy-pasted from the old softras example

examples/renderers/softras_vertex_optimization.py
krrish94: Good catch, this was copy-pasted from the old softras example

examples/renderers/softras_vertex_optimization.py
krrish94: We already have something inside transforms that's maybe close. Will be nice to have a full-fledged feature

examples/renderers/softras_vertex_optimization.py
krrish94: this was copy-pasted from the old softras example; will change

kaolin/graphics/Lighting.py
krrish94: This convention is tweaked (as opposed to using our original), for exactness wrt softras. (In fact our first release used our function from `TriangleMesh`)

kaolin/graphics/SoftRenderer.py
krrish94: I think it does. Eg. people can often mistakenly pass `distance=2` when they mean 2 meters, or `azimuth=90` when they mean 90 degrees

tests/graphics/test_softras.py
krrish94: this was copy-pasted from the old softras example; will remove

tests/graphics/test_softras.py
krrish94: Good catch, this was copy-pasted from the old softras example

tests/graphics/test_softras.py
krrish94: Good catch, this was copy-pasted from the old softras example

krrish94:
Thanks for the (very) useful review comments! I see two major issues:
1. Where do we store these examples: I agree with you that we should move these into `examples/renderers/softras`.
2. Where do we store the data for the examples (`banana.png`, `banana.obj`, `model.obj`)? I think we should store them inside `examples/renderers/sampledata` to allow for reuse across renderers (like NMR and DIB-R)

Waiting to hear what you have to say about (2).

kaolin/graphics/SoftRenderer.py
Caenorst: Sorry, what I mean is explicitly saying that we will support int, user should understand that the input will be interpreted as a float (if it can).

For instance https://pytorch.org/docs/stable/torch.html#torch.pow only says `float or tensor` but in fact `int` works just fine. It make the doc lighter.

Caenorst:
storing those sample in `examples/renderers/sampledata` make sense

kaolin/graphics/SoftRenderer.py
krrish94: Makes sense

Caenorst:
pro-tips: using `git mv` will move the file, and then git will keep track of the history of the file.
And of my comments too.

kaolin/graphics/Lighting.py
Caenorst: Yes, but it's exactly the same than the one for MeshCNN then, right ?

Here is a suggestion, let add an option to https://github.com/NVIDIAGameWorks/kaolin/blob/master/kaolin/rep/TriangleMesh.py#L154
like `robust` or `averaged` (or whatever argument name that would indicate that you average over all cross products instead of picking just one).

It would else for reusability and would avoid out-of-the-core calculation on MeshCNN.

krrish94: Yeah, I think that'll be great. We'd want these to be "inside core" and standardized to the extent possible

Caenorst (approved):
LGTM, thanks for the contrib ! :)

krrish94:
Closes #236

PR #200, opened 2020-04-14
Fix/msvc compilation

radekd91:
This PR fixes two issues with compilation on Windows using MSVC. 
- forward declaration inconsistency with the actual definition of the cuda function `MeshIntersectionKernelLauncher` 
- fix of preprocessor condition syntax that was not working properly on MSVC

With these two fixes, I was able to compile on Windows just fine, the setup was the following:
Python version: 3.6
CUDA version: 10.1
Pytorch version: 1.4.0

PR #238, opened 2020-05-01
windows build fixes

jasongorski:
- don't include pytorch in some .cu to avoid constexpr and cast errors from nvcc
- function prototype and preprocessor fixes (some overlap with https://github.com/NVIDIAGameWorks/kaolin/pull/200)
- exclude gcc/clang compiler flags from windows build


Caenorst (dismissed):
LGTM

Caenorst:
little conflict in .gitlab-ci.yml since we add some more tests

Caenorst (approved):

PR #248, opened 2020-05-05
drop install of gtk from CI steps

jasongorski:
drop libgtk2.0-dev .deb requirement by:

- switching from opencv-python to opencv-python-headless
- switching matplotlib default backend away from one requiring interactive x11 session


krrish94 (commented):
LGTM

krrish94 (approved):

PR #254, opened 2020-05-08
add nuscenes-devkit to installation requirement

ethanjyx:
Signed-off-by: Yixing Jiang <ethanjyx@umich.edu>

This library should be required, otherwise I'm getting this error
```
>>> import kaolin as kal
Warning: unable to import datasets/nusc:
   No module named 'nuscenes'
Traceback (most recent call last):
  File "/nfs1/jyx/kaolin/kaolin/datasets/__init__.py", line 11, in <module>
    from .nusc import NuscDetection
  File "/nfs1/jyx/kaolin/kaolin/datasets/nusc.py", line 21, in <module>
    from nuscenes.utils.geometry_utils import transform_matrix
ModuleNotFoundError: No module named 'nuscenes'
Warning: unable to import datasets/nusc:
   None
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -209,6 +209,7 @@ cwd = os.path.dirname(os.path.abspath(__file__))
 def get_requirements():
     return [
         'matplotlib<3.0.0',
+        'nuscenes-devkit',
         'scikit-image',
         'shapely',
         'trimesh>=3.0',


Caenorst:
Hi @ethanjyx ! Thank you for your interest in Kaolin.

The message is actually a warning message. We don't want to force user to install nuscenes-devkit and all the dependencies if you are not using the dataset so we prefer to keep it this way.

We probably would need to change the warning message to be more explicit as a warning and not an error.

shumash:
How about using: `importlib.util.find_spec` to produce a more friendly warning?

Issue #250, opened 2020-05-06
TranslatePointCloud transform


pushkalkatara:
Adding a new transform function in kaolin.Transform : translatePointCloud.

In dgcnn, author have used it as a preprocessing function: https://github.com/WangYueFt/dgcnn/blob/20fdb459ca5d10fe8aba1d296e66340f65990b85/pytorch/data.py#L50

pushkalkatara:
The preprocessing function is 
```
(pc + xyz1) * xyz2
```

The mulitplication operation on pointcloud can be performed through https://github.com/NVIDIAGameWorks/kaolin/blob/c3ec5484c326e83d6e1a9fdfd170b4c67343c68b/kaolin/transforms/transforms.py#L201
but the addition operation is missing. Adding translate as a transform function might be useful

krrish94:
Thanks for opening this issue! This would be a nice addition, and relatively simple to implement.

SomTambe:
Can I work on this issue ? This seems interesting to implement 😀

SomTambe:
I aim to do this by the following-
- Add a translate function to kaolin.transforms.pointcloudfunc
- Add the translatePointCloud  transform to kaolin.transforms.transforms, by making use of ScalePointCloud transforms also.

Could you assign this to me ?

shumash:
Closing old issue, as it is no longer relevant to the library.

PR #255, opened 2020-05-08
Translate transform for pointclouds

pushkalkatara:
This PR solves issue #250

### DIFF
--- a/kaolin/transforms/pointcloudfunc.py
+++ b/kaolin/transforms/pointcloudfunc.py
@@ -26,17 +26,16 @@ EPS = 1e-6
 
 
 def shift(cloud: Union[torch.Tensor, PointCloud],
-          shf: Union[float, int, torch.Tensor],
-          inplace: Optional[bool] = True):
+          shf: Union[float, int, torch.Tensor]):
     """Shift the input pointcloud by a shift factor.
 
     Args:
         cloud (torch.Tensor or kaolin.rep.PointCloud): pointcloud (ndims >= 2).
-        shf (float, int, torch.Tensor): shift factor (scaler, or tensor).
+        shf (float, int, torch.Tensor): shift factor (scalar, or tensor).
         inplace (bool, optional): Bool to make the transform in-place
     
     Returns:
-        (torch.Tensor): shifted pointcloud pf the same shape as input.
+        (torch.Tensor): shifted pointcloud of the same shape as input.
     
     Shape:
         - cloud: :math:`(B x N x D)` (or) :math:`(N x D)`, where :math:`(B)`
@@ -48,25 +47,15 @@ def shift(cloud: Union[torch.Tensor, PointCloud],
         >>> points = torch.rand(1000,3)
         >>> points2 = shift(points, torch.FloatTensor([3]))
     """
-
-    if isinstance(cloud, np.ndarray):
-        cloud = torch.from_numpy(cloud)
-    
-    if isinstance(shf, np.ndarray):
-        shf = torch.from_numpy(shf)
-    
     if isinstance(cloud, PointCloud):
         cloud = cloud.points
     
     if isinstance(shf, int) or isinstance(shf, float):
-        shf = torch.Tensor([shf]).to(cloud.device)
+        shf = torch.Tensor([shf], device=cloud.device)
     
     helpers._assert_tensor(cloud)
     helpers._assert_tensor(shf)
     helpers._assert_dim_ge(cloud, 2)
-
-    if not inplace:
-        cloud = cloud.clone()
 
     return shf + cloud
 
@@ -96,18 +85,11 @@ def scale(cloud: Union[torch.Tensor, PointCloud],
         >>> points2 = scale(points, torch.FloatTensor([3]))
 
     """
-
-    if isinstance(cloud, np.ndarray):
-        cloud = torch.from_numpy(cloud)
-
-    if isinstance(scf, np.ndarray):
-        scf = torch.from_numpy(scf)
-
     if isinstance(cloud, PointCloud):
         cloud = cloud.points
 
     if isinstance(scf, int) or isinstance(scf, float):
-        scf = torch.Tensor([scf]).to(cloud.device)
+        scf = torch.Tensor([scf], device=cloud.device)
 
     helpers._assert_tensor(cloud)
     helpers._assert_tensor(scf)
@@ -120,16 +102,15 @@ def scale(cloud: Union[torch.Tensor, PointCloud],
     return scf * cloud
 
 
-def translate(cloud: Union[torch.Tensor, PointCloud], tranmat: torch.Tensor,
-              inplace: Optional[bool] = True):
+def translate(cloud: Union[torch.Tensor, PointCloud], tranmat: torch.Tensor):
     """Translate the input pointcloud by a translation matrix.
 
     Args:
-        cloud (Tensor or np.array): pointcloud (ndims = 2 or 3)
-        tranmat (Tensor or np.array): translation matrix (1 x 3, 1 per cloud).
+        cloud (torch.Tensor or kaolin.rep.PointCloud): pointcloud (ndims = 2 or 3)
+        tranmat (torch.Tensor or np.array): translation matrix (1 x 3, 1 per cloud).
 
     Returns:
-        cloud_tran (Tensor): Translated pointcloud of the same shape as input.
+        (torch.Tensor): Translated pointcloud of the same shape as input.
 
     Shape:
         - cloud: :math:`(B x N x 3)` (or) :math:`(N x 3)`, where :math:`(B)`
@@ -143,8 +124,6 @@ def translate(cloud: Union[torch.Tensor, PointCloud], tranmat: torch.Tensor,
         >>> points2 = translate(points, t_mat)
 
     """
-    if isinstance(cloud, np.ndarray):
-        cloud = torch.from_numpy(cloud)
     if isinstance(cloud, PointCloud):
         cloud = cloud.points
     if isinstance(tranmat, np.ndarray):
@@ -154,33 +133,24 @@ def translate(cloud: Union[torch.Tensor, PointCloud], tranmat: torch.Tensor,
     helpers._assert_tensor(tranmat)
     helpers._assert_dim_ge(cloud, 2)
     helpers._assert_dim_ge(tranmat, 2)
-    # Rotation matrix must have last two dimensions of shape 3.
-    helpers._assert_shape_eq(tranmat, (1, 3), dim=-1)
-    helpers._assert_shape_eq(tranmat, (1, 3), dim=-2)
+    helpers._assert_dim_le(cloud, 3)
+    helpers._assert_dim_le(tranmat, 3)
+    # Translation matrix and cloud last dimension must be equal
+    helpers._assert_shape_eq(tranmat, cloud.shape, dim=-1)
 
-    if not inplace:
-        cloud = cloud.clone()
-
-    if tranmat.dim() == 2 and cloud.dim() == 2:
-        cloud = torch.add(tranmat, cloud)
-    else:
-        if tranmat.dim() == 2:
-            tranmat = tranmat.expand(cloud.shape[0], 1, 3)
-        cloud = torch.add(tranmat, cloud)
-
-    return cloud
+    return torch.add(tranmat, cloud)
 
 def rotate(cloud: Union[torch.Tensor, PointCloud], rotmat: torch.Tensor,
            inplace: Optional[bool] = True):
     """Rotates the the input pointcloud by a rotation matrix.
 
     Args:
-        cloud (Tensor or np.array): pointcloud (ndims = 2 or 3)
-        rotmat (Tensor or np.array): rotation matrix (3 x 3, 1 per cloud).
+        cloud (torch.Tensor or kaolin.rep.PointCloud): pointcloud (ndims = 2 or 3)
+        rotmat (torch.Tensor or np.array): rotation matrix (3 x 3, 1 per cloud).
         inplace (bool, optional): Bool to make the transform in-place.
 
     Returns:
-        cloud_rot (Tensor): rotated pointcloud of the same shape as input
+        (torch.Tensor): rotated pointcloud of the same shape as input
 
     Shape:
         - cloud: :math:`(B x N x 3)` (or) :math:`(N x 3)`, where :math:`(B)`
@@ -194,8 +164,6 @@ def rotate(cloud: Union[torch.Tensor, PointCloud], rotmat: torch.Tensor,
         >>> points2 = rotate(points, r_mat)
 
     """
-    if isinstance(cloud, np.ndarray):
-        cloud = torch.from_numpy(cloud)
     if isinstance(cloud, PointCloud):
         cloud = cloud.points
     if isinstance(rotmat, np.ndarray):
@@ -229,15 +197,15 @@ def realign(src: Union[torch.Tensor, PointCloud],
     box as that of pointcloud `tgt`.
 
     Args:
-        src (torch.Tensor or PointCloud) : Source pointcloud to be transformed
-            (shape: :math:`\cdots \times N \times D`, where :math:`N` is the
-            number of points in the pointcloud, and :math:`D` is the
-            dimensionality of each point in the cloud).
-        tgt (torch.Tensor or PointCloud) : Target pointcloud to which `src`is
-            to be transformed (The `src` cloud is transformed to the
-            axis-aligned bounding box that the target cloud maps to). This
-            cloud must have the same number of dimensions :math:`D` as in the
-            source cloud. (shape: :math:`\cdots \times \cdots \times D`).
+        src ((torch.Tensor or kaolin.rep.PointCloud)) : Source pointcloud to
+             be transformed (shape: :math:`\cdots \times N \times D`, where 
+             :math:`N` is the number of points in the pointcloud, and :math:`D`
+             is the dimensionality of each point in the cloud).
+        tgt ((torch.Tensor or kaolin.rep.PointCloud)) : Target pointcloud to 
+            which `src`is to be transformed (The `src` cloud is transformed 
+            to the axis-aligned bounding box that the target cloud maps to). 
+            This cloud must have the same number of dimensions :math:`D` as 
+            in the source cloud. (shape: :math:`\cdots \times \cdots \times D`).
         inplace (bool, optional): Bool to make the transform in-place.
 
     Returns:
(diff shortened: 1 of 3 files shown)


pushkalkatara:
The build is failing at weird error - 
```bash

/opt/conda/lib/python3.6/site-packages/numpy/core/include/numpy/npy_1_7_deprecated_api.h:17:2: warning: #warning "Using deprecated NumPy API, disable it with " "#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION" [-Wcpp]
--
831 | #warning "Using deprecated NumPy API, disable it with " \
832 | ^
833 | imagecodecs/_aec.c:602:20: fatal error: libaec.h: No such file or directory
834 | compilation terminated.
835 | error: Setup script exited with error: command 'gcc' failed with exit status 1
```
maybe adding `apt-get install libaec-dev` would solve build error

pushkalkatara:
Hi @krrish94 , @Caenorst This transform is a preprocessing function for semantic segmentation in DGCNN pipeline. Please review the PR.

Caenorst:
Hi @pushkalkatara , I think the error that you faced might be due to recent numpy release, we fixed it in #262 can you rebase and force push ?

pushkalkatara:
Thanks for pointing out @Caenorst Now the build is passing :D

kaolin/transforms/pointcloudfunc.py
```diff
@@ -25,6 +25,52 @@
...
+    
+    if isinstance(shf, int) or isinstance(shf, float):
+        shf = torch.Tensor([shf]).to(cloud.device)
+    
+    helpers._assert_tensor(cloud)
+    helpers._assert_tensor(shf)
+    helpers._assert_dim_ge(cloud, 2)
+
+    if not inplace:
+        cloud = cloud.clone()
```
Caenorst: This is not done inplace, `shf + cloud` return a new tensor, there is a way to make it inplace but that would break differentiability. I would suggest to just remove the option.

kaolin/transforms/pointcloudfunc.py
```diff
@@ -25,6 +25,52 @@
...
+            is the batchsize, :math:`(N)` is the number of points per cloud,
+            and :math:`(D)` is the dimensionality of each cloud.
+        - shf: :math:`(1)` or :math:`(B)`.
+
+    Example:
+        >>> points = torch.rand(1000,3)
+        >>> points2 = shift(points, torch.FloatTensor([3]))
+    """
+
+    if isinstance(cloud, np.ndarray):
```
Caenorst: This is not consistent with documentation and type hints `torch.Tensor` or `kaolin.rep.PointCloud` we can drop the numpy support and assume the user will do the conversion.

kaolin/transforms/pointcloudfunc.py
```diff
@@ -25,6 +25,52 @@
...
+        cloud = torch.from_numpy(cloud)
+    
+    if isinstance(shf, np.ndarray):
```
Caenorst: Same this is not consistent with documentation and type hints

kaolin/transforms/pointcloudfunc.py
```diff
@@ -25,6 +25,52 @@
...
+        shf = torch.from_numpy(shf)
+    
+    if isinstance(cloud, PointCloud):
+        cloud = cloud.points
+    
+    if isinstance(shf, int) or isinstance(shf, float):
+        shf = torch.Tensor([shf]).to(cloud.device)
```
Caenorst: torch support addition of `torch.Tensor` with python `int` or `float`
note: `shf = torch.tensor([shf], device=cloud.device)` is better (avoid transfer to GPU)
Also what happens if shf is numpy array (or torch.cpu.Tensor) and cloud is torch.cuda.Tensor ?

kaolin/transforms/pointcloudfunc.py
```diff
@@ -74,6 +120,56 @@ def scale(cloud: Union[torch.Tensor, PointCloud],
...
+    
+    helpers._assert_tensor(cloud)
+    helpers._assert_tensor(tranmat)
+    helpers._assert_dim_ge(cloud, 2)
+    helpers._assert_dim_ge(tranmat, 2)
+    # Rotation matrix must have last two dimensions of shape 3.
+    helpers._assert_shape_eq(tranmat, (1, 3), dim=-1)
+    helpers._assert_shape_eq(tranmat, (1, 3), dim=-2)
+
+    if not inplace:
```
Caenorst: same remark than for shift, I think `inplace` is not doing anything

kaolin/transforms/pointcloudfunc.py
```diff
@@ -25,6 +25,52 @@
...
+          inplace: Optional[bool] = True):
+    """Shift the input pointcloud by a shift factor.
+
+    Args:
+        cloud (torch.Tensor or kaolin.rep.PointCloud): pointcloud (ndims >= 2).
+        shf (float, int, torch.Tensor): shift factor (scaler, or tensor).
+        inplace (bool, optional): Bool to make the transform in-place
+    
+    Returns:
+        (torch.Tensor): shifted pointcloud pf the same shape as input.
```
Caenorst: typo: shifted pointcloud **o**f the same shape as input.

kaolin/transforms/pointcloudfunc.py
```diff
@@ -74,6 +120,56 @@ def scale(cloud: Union[torch.Tensor, PointCloud],
...
+def translate(cloud: Union[torch.Tensor, PointCloud], tranmat: torch.Tensor,
+              inplace: Optional[bool] = True):
+    """Translate the input pointcloud by a translation matrix.
+
+    Args:
+        cloud (Tensor or np.array): pointcloud (ndims = 2 or 3)
+        tranmat (Tensor or np.array): translation matrix (1 x 3, 1 per cloud).
+
+    Returns:
+        cloud_tran (Tensor): Translated pointcloud of the same shape as input.
```
Caenorst: Returned values should not be named in the documentation.

kaolin/transforms/pointcloudfunc.py
```diff
@@ -74,6 +120,56 @@ def scale(cloud: Union[torch.Tensor, PointCloud],
     return scf * cloud
 
 
+def translate(cloud: Union[torch.Tensor, PointCloud], tranmat: torch.Tensor,
+              inplace: Optional[bool] = True):
+    """Translate the input pointcloud by a translation matrix.
+
+    Args:
+        cloud (Tensor or np.array): pointcloud (ndims = 2 or 3)
```
Caenorst: This is not consistent with the type hints (`torch.Tensor` or `kaolin.rep.PointCloud`).
`torch.Tensor` should be preferred over `Tensor`

kaolin/transforms/pointcloudfunc.py
```diff
@@ -74,6 +120,56 @@ def scale(cloud: Union[torch.Tensor, PointCloud],
...
+        tranmat (Tensor or np.array): translation matrix (1 x 3, 1 per cloud).
```
Caenorst: This is not consistent with the type hints (`torch.Tensor`).
Also `torch.Tensor` should be preferred over `Tensor`.

kaolin/transforms/pointcloudfunc.py
```diff
@@ -74,6 +120,56 @@ def scale(cloud: Union[torch.Tensor, PointCloud],
...
+
+    Example:
+        >>> points = torch.rand(1000,3)
+        >>> t_mat = torch.rand(1,3)
+        >>> points2 = translate(points, t_mat)
+
+    """
+    if isinstance(cloud, np.ndarray):
+        cloud = torch.from_numpy(cloud)
+    if isinstance(cloud, PointCloud):
```
Caenorst, line 130: this is not consistent with the documentation (but is consistent with type hints).

kaolin/transforms/pointcloudfunc.py
```diff
@@ -74,6 +120,56 @@ def scale(cloud: Union[torch.Tensor, PointCloud],
...
+        cloud = cloud.points
+    if isinstance(tranmat, np.ndarray):
+        trainmat = torch.from_numpy(tranmat)
+    
+    helpers._assert_tensor(cloud)
+    helpers._assert_tensor(tranmat)
+    helpers._assert_dim_ge(cloud, 2)
+    helpers._assert_dim_ge(tranmat, 2)
```
Caenorst, line 138: those check that the dim is greater or equal to 2 but not smaller than 4 (which seems to be assumed by the documentation ?)

kaolin/transforms/pointcloudfunc.py
```diff
@@ -74,6 +120,56 @@ def scale(cloud: Union[torch.Tensor, PointCloud],
...
+    helpers._assert_shape_eq(tranmat, (1, 3), dim=-1)
+    helpers._assert_shape_eq(tranmat, (1, 3), dim=-2)
+
+    if not inplace:
+        cloud = cloud.clone()
+
+    if tranmat.dim() == 2 and cloud.dim() == 2:
+        cloud = torch.add(tranmat, cloud)
+    else:
+        if tranmat.dim() == 2:
```
Caenorst: what if `cloud.dim() == 2` and `transmat.dim() == 3` ?

kaolin/transforms/transforms.py
```diff
@@ -231,6 +260,35 @@ def __call__(self, cloud: Union[torch.Tensor, PointCloud]):
...
+    r"""Translate a pointcloud with a given translation matrix.
+    Given a :math:`1 \times 3` translation matrix, this transform will 
+    translate each point in the cloud by the translation matrix specified.
+
+    Args:
+        tranmat (torch.Tensor): Translation matrix that specifies the translation 
+            to be applied to the pointcloud (shape: :math:`1 \times 3`).
+        inplace (bool, optional): Bool to make this operation in-place.
+
+    TODO: Example.
```
Caenorst: can you add the example ?

kaolin/transforms/pointcloudfunc.py
```diff
@@ -25,6 +25,52 @@
...
 
 
+def shift(cloud: Union[torch.Tensor, PointCloud],
+          shf: Union[float, int, torch.Tensor],
+          inplace: Optional[bool] = True):
+    """Shift the input pointcloud by a shift factor.
+
+    Args:
+        cloud (torch.Tensor or kaolin.rep.PointCloud): pointcloud (ndims >= 2).
+        shf (float, int, torch.Tensor): shift factor (scaler, or tensor).
```
Caenorst: typo: scal**a**r

kaolin/transforms/pointcloudfunc.py
```diff
@@ -25,6 +25,52 @@
...
+    
+    if isinstance(cloud, PointCloud):
+        cloud = cloud.points
+    
+    if isinstance(shf, int) or isinstance(shf, float):
+        shf = torch.Tensor([shf]).to(cloud.device)
+    
+    helpers._assert_tensor(cloud)
+    helpers._assert_tensor(shf)
+    helpers._assert_dim_ge(cloud, 2)
```
Caenorst, line 59: There is no test for cloud number of dims < 4

tests/transforms/test_transforms.py
```diff
@@ -38,6 +45,14 @@ def test_scale_pointcloud(device='cpu'):
     assert_allclose(pc_, torch.ones(4, 3))
 
 
+def test_translate_pointcloud(device='cpu'):
```
Caenorst, line 48: can you test more variation of dimensions ? (2 and 3)

Caenorst (changes requested):
Thank you for your interest and contribution for Kaolin !

There are some comments that will need to be addressed, otherwise it's looking good ! :)

kaolin/transforms/pointcloudfunc.py
pushkalkatara: Does inplace for `scale` transform operation also not work?

we can add inplace option if required : https://discuss.pytorch.org/t/what-is-in-place-operation/16244

kaolin/transforms/pointcloudfunc.py
pushkalkatara: nice catch! thanks for pointing out.

kaolin/transforms/pointcloudfunc.py
pushkalkatara: Adding in next commit, thanks for pointing out!

kaolin/transforms/pointcloudfunc.py
pushkalkatara: Thanks for pointing out! I'll update it in next commit.

kaolin/transforms/pointcloudfunc.py
pushkalkatara: Should this be updated to kaolin.rep.PointCloud ?

kaolin/transforms/pointcloudfunc.py
pushkalkatara: Adding one more assert to check le=3

kaolin/transforms/pointcloudfunc.py
pushkalkatara: Ideally, last dimension of the tensor for cloud and tranmat should be equal. If its i 3D point cloud, user should give provide a 3D tranmat. else if it is a 2D pointcloud, user should provide 2D tranmat. 

I this is the right strategy, we can add an assert checking the last dimesion is equal for both tensors, cloud and tranmat.

kaolin/transforms/transforms.py
pushkalkatara: Added in next commit :D

tests/transforms/test_transforms.py
pushkalkatara: Added more tests in next commit :D

pushkalkatara:
Hi @Caenorst Thanks for the detailed review. I have made some changes in the recent commit and added questions as well. please review once again. Thanks.

kaolin/transforms/pointcloudfunc.py
Kh4L: I believe that if `cloud` doesn't require grads, having an inplace add would be fine here:
```
if inplace:
    cloud.add_(shf)
    return
```

@Caenorst to verify if relying on torch's autograd runtime error for Transforms is something we want here.

kaolin/transforms/pointcloudfunc.py
Caenorst: My note was actually not a strong indication for the PR (more like a general advice :) )

I believe converting shf to torch.Tensor is not useful as https://github.com/pushkalkatara/kaolin/blob/translatePC/kaolin/transforms/pointcloudfunc.py#L60 would work just as fine with python `int` and `float`. It would work regardless of `cloud.device`.

kaolin/transforms/pointcloudfunc.py
Caenorst: Unresolving: the documentation still indicate that tranmat can be `torch.Tensor` or `np.array` while the type hints is indicating `torch.Tensor` only

kaolin/transforms/pointcloudfunc.py
Caenorst: I think referencing to PointCloud is fine but being more specific won't hurt :)

kaolin/transforms/pointcloudfunc.py
```diff
@@ -25,6 +25,41 @@
...
+    
+    Shape:
+        - cloud: :math:`(B x N x D)` (or) :math:`(N x D)`, where :math:`(B)`
+            is the batchsize, :math:`(N)` is the number of points per cloud,
+            and :math:`(D)` is the dimensionality of each cloud.
+        - shf: :math:`(1)` or :math:`(B)`.
+
+    Example:
+        >>> points = torch.rand(1000,3)
+        >>> points2 = shift(points, torch.FloatTensor([3]))
```
Caenorst: I would suggest an example that is more explicit about what it is doing, for instance:
```
>>> points = torch.tensor([[0., 0., 0.],
                                          [1., 0., 0.],
                                          [2., 3., 4.]])
>>> shift(points, 3)
tensor([3., 3., 3.],
           [4., 3., 3.],
           [5., 6., 7.]])
```

kaolin/transforms/pointcloudfunc.py
```diff
@@ -74,17 +102,55 @@ def scale(cloud: Union[torch.Tensor, PointCloud],
...
+    Shape:
+        - cloud: :math:`(B x N x 3)` (or) :math:`(N x 3)`, where :math:`(B)`
+            is the batchsize, :math:`(N)` is the number of points per cloud,
+            and :math:`(3)` is the dimensionality of each cloud.
+        - tranmat: :math:`(1, 3)` or :math:`(B, 1, 3)`.
+
+    Example:
+        >>> points = torch.rand(1000,3)
+        >>> t_mat = torch.rand(1,3)
+        >>> points2 = translate(points, t_mat)
```
Caenorst, line 127: same remark than for shift

Caenorst (commented):
a little comment about examples

kaolin/transforms/pointcloudfunc.py
pushkalkatara: Sure, I'll make the changes!

kaolin/transforms/pointcloudfunc.py
pushkalkatara: I'll make the change here as well.

kaolin/transforms/pointcloudfunc.py
pushkalkatara: Oh missed it. I'll update in the next commit.

kaolin/transforms/pointcloudfunc.py
pushkalkatara: Other functions as well like rotate were using inplace. What change shall I go forward with?

Kh4L: @Caenorst for final decision

pushkalkatara: Then, shall I remove inplace for other functions as well?

Caenorst: We can take care of that on separate PR :)

kaolin/transforms/pointcloudfunc.py
Caenorst: Can you just show the output of `translate(points, t_mat)` ?
```
>>> t_mat = torch.tensor([0.2, 0.3 ,0.4])
>>> translate(points, t_mat)
tensor([[0.2, 0.3, 0.4].
            [1.2, 2.2, 3.3],
            [2.1, 3.2, 4.3]])
```
(if I'm correct with the op)
The role of example is to show how it's used and what it does, if you don't show the output depending on input it's very hard for the user to know what it does.

kaolin/transforms/pointcloudfunc.py
pushkalkatara: Okay sure :D Is this PR up to the mark?

Caenorst:
Sorry, forgot to confirm the review :/

shumash:

PR #278, opened 2020-06-18
Update STYLE_GUIDE.md

sambhav228:
Reading Document should be effective for reader/users


shumash:

PR #345, opened 2020-12-18
ShapeNet dataset

TommyX12:
- Implemented KaolinDataset base class to automatically combine the return values of `get_data` and `get_attributes` in `__getitem__`
- Implemented ShapeNet dataset, using `kaolin.io.obj.import_mesh`

Tested on Ubuntu 18.04, PyTorch 1.7.0, Python 3.6.9.


kaolin/datasets/shapenet.py
```diff
@@ -0,0 +1,147 @@
+# Copyright (c) 2019-2020, NVIDIA CORPORATION. All rights reserved.
```
Caenorst: I think this should be in `kaolin/io/shapenet.py`, it's better not to add another namespace at the root directory

kaolin/io/obj.py
```diff
@@ -156,15 +164,17 @@ def import_mesh(path, with_materials=False, with_normals=False, error_handler=No
...
+            [float(el) for sublist in vertex_normals
+             for el in sublist]).view(-1, 3)
         face_normals = torch.LongTensor(face_normals) - 1
     else:
         vertex_normals = None
         face_normals = None
 
-    return return_type(vertices, faces, uvs, face_uvs_idx, materials,
-                       materials_order, vertex_normals, face_normals)
+    return ObjMesh(vertices, faces, uvs, face_uvs_idx, materials,
```
Caenorst, line 175: AFAIK nametuple shouldn't start with a capital letter.

Also I think we should keep it as `kaolin.io.obj.return_type`. it is pretty standard in pytorch, see https://pytorch.org/docs/stable/generated/torch.topk.html or https://pytorch.org/docs/stable/generated/torch.max.html ...

Will see later when we have a high level rep :)

tests/python/kaolin/datasets/test_shapenet.py
```diff
@@ -0,0 +1,60 @@
...
+        assert isinstance(attributes['synset'], str)
+        assert isinstance(attributes['label'], str)
+
+    def test_init_by_labels(self):
+        dataset = ShapeNet(root=SHAPENET_PATH,
+                           categories=SHAPENET_TEST_CATEGORY_LABELS,
+                           train=True, split=1.0)
+
+        assert isinstance(dataset[0].data, ObjMesh)
+        assert isinstance(dataset[0].attributes, dict)
```
Caenorst: So, this is gonna require a little bit more testing:
1) Test different split and categories
2) You can test for size and get the value at an index (it should be deterministic, but the value depend if we are using true or dummy shapenet)

TommyX12: I'm not sure how many examples will there be in the dummy shapenet, so i didn't test for the split. What would be the preferred way to know if im in the dummy dataset?

kaolin/datasets/shapenet.py
TommyX12: Moved!

kaolin/io/obj.py
TommyX12: I followed the CamelCase convention since it's acts as a factory function: https://stackoverflow.com/questions/36371047/python-naming-convention-namedtuples

Also I think PyTorch used `return_types` since the return type of these operations are usually "anonymous" to the outside. However, I think the return type of `import_mesh` deserves its own meaningful name since it represents concrete data (i.e. a mesh) instead of just the result of some operation. As a user I would also want to import the name of this mesh "class" to annotate types or perform assertion, and it would make a bit more sense if I import `kaolin.io.obj.ObjMesh` instead of `kaolin.io.obj.return_type as obj_return_type`.

To truly follow PyTorch's convention we would also have to make it `kaolin.return_types.import_mesh` instead; there might be more than one `return_type` needed in `obj.py`.

What do you think?

tests/python/kaolin/datasets/test_shapenet.py
TommyX12: Update: added more tests

Caenorst: Using an environment variable would be ideal I think

kaolin/io/obj.py
Caenorst: I hear you, however we should stay consistent with https://github.com/NVIDIAGameWorks/kaolin/blob/master/kaolin/io/usd.py `mesh_return_type`. Also this is supposed to be anonymous, we don't want users to use this nametuple to create their own (and it's not necessary to import the type to run assertion), this is the output of the obj loader, not a standard mesh representation (which will be implemented later on, if we start to have this nametuple as a standard rep then it will be confused with the higher rep later), I think it's better to keep it as `return_type` or `obj_return_type`. There should be no multiple return_type from obj loader.

tests/python/kaolin/datasets/test_shapenet.py
Caenorst: Can you add some test for values of output (select a fixed index for a given instance of the dataset you should always have the same output).

Caenorst:
Closing as we took over in #373

PR #346, opened 2021-01-06
remove compare error string in test_obj.py

JerryJiehanWang:
test_obj.py's string comparison will result in error in windows, thus remove it.


Caenorst (approved):
LGTM ! :)

PR #356, opened 2021-02-23
Replace os.path.join by posixpath.join in all usd exporting functions and update posix texture path.

Caenorst:
Fix some issues with windows


Jean-Francois-Lafleche (approved):
LGTM!

PR #357, opened 2021-02-25
Fix documentation issues

Caenorst:
Generated documentation can be checked on: https://kaolin--357.org.readthedocs.build/en/357/

We still need to find a way to make the build explicitly fail when those warnings happens.

Signed-off-by: Clement Fuji Tsang <cfujitsang@nvidia.com>


JerryJiehanWang (approved):
LGTM!

PR #361, opened 2021-03-12
Pointcloud to voxelgrid conversion

JerryJiehanWang:
Pointcloud to voxelgrid conversion

### DIFF
--- a/kaolin/ops/conversions/pointcloud.py
+++ b/kaolin/ops/conversions/pointcloud.py
@@ -18,8 +18,7 @@ __all__ = ['pointclouds_to_voxelgrids']
 
 def _base_points_to_voxelgrids(points, resolution, return_sparse=False):
     r"""Converts points to voxelgrids. This is the base function for both trianglemeshes_to_voxelgrids
-    and pointclouds_to_voxelgrids. For point cloud, the points are the original points. For Triangle Mesh,
-    the points are sampled along the edges. The points are normalized into range [0, 1].
+    and pointclouds_to_voxelgrids. The points are normalized into range [0, 1].
 
     Args:
         points (torch.Tensor):
--- a/kaolin/ops/conversions/pointcloud.py
+++ b/kaolin/ops/conversions/pointcloud.py
@@ -16,6 +16,56 @@ import torch
 
 __all__ = ['pointclouds_to_voxelgrids']
 
+def _points_to_voxelgrids(points, resolution, return_sparse=False):
+    r"""Converts points to voxelgrids. This is the base function for both trianglemeshes_to_voxelgrids
+    and pointclouds_to_voxelgrids.
+
+    Args:
+        points (torch.Tensor):
+            Exact batched points with shape
+            :math: `(\text{batch_size}, \text{P}, \text{3})
+        resolution (int):
+            Resolution of output voxelgrids
+        return_sparse (bool):
+            Whether to return a sparse voxelgrids or not.
+
+    Returns:
+        (torch.Tensor or torch.FloatTensor):
+        Exact batched voxelgrids with shape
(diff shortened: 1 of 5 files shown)


kaolin/ops/conversions/pointcloud.py
```diff
@@ -0,0 +1,83 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+
+__all__ = ['pointclouds_to_voxelgrids']
+
+def pointclouds_to_voxelgrids(pointclouds, resolution, return_sparse=False):
```
Caenorst: shouldn't we adopt the same interface than `trianglemesh_to_voxelgrids` ? https://github.com/NVIDIAGameWorks/kaolin/blob/master/kaolin/ops/conversions/trianglemesh.py#L20

JerryJiehanWang: Do you mean we should also add origin and scale as arguments?

Caenorst: Yes!

JerryJiehanWang: Added, as well as the tests.

kaolin/ops/conversions/pointcloud.py
```diff
@@ -0,0 +1,129 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+
+__all__ = ['pointclouds_to_voxelgrids']
+
+def _points_to_voxelgrids(points, resolution, return_sparse=False):
```
Caenorst: maybe add the prefix `_base` like for: https://github.com/NVIDIAGameWorks/kaolin/blob/master/kaolin/ops/mesh/trianglemesh.py#L26

Also maybe we can add some details about the behavior? (i.e: For what I understand, it only convert points from 0 to 1)

kaolin/ops/conversions/pointcloud.py
```diff
@@ -0,0 +1,130 @@
...
+# limitations under the License.
+
+import torch
+
+__all__ = ['pointclouds_to_voxelgrids']
+
+def _base_points_to_voxelgrids(points, resolution, return_sparse=False):
+    r"""Converts points to voxelgrids. This is the base function for both trianglemeshes_to_voxelgrids
+    and pointclouds_to_voxelgrids. For point cloud, the points are the original points. For Triangle Mesh,
+    the points are sampled along the edges. The points are normalized into range [0, 1].
```
Caenorst: I'm not sure we should reference mesh behavior, the behavior that you described here ("For Triangle Meshes, the points are samples along the edges") has nothing to do with this function, isn't it? If you changed the way you sampled points along the edges this function would still be the same the description would be false, which is difficult to maintain.

Also, are the points normalized? It seemed to be that only points in the range [0, 1] (before doing anything to them) were sampled.

JerryJiehanWang: Yes, the points are normalized. So will it be sufficient to say that the function converts normalized points in the range [0, 1] to voxelgrid?

Caenorst: My understanding is that the point cloud MAY not be normalized, but regardless of that this base function will ONLY voxelize points that are in the range [0, 1]. Am I correct?

Caenorst (approved):
LGTM, Thanks!

PR #373, opened 2021-04-12
ShapeNet Dataset

JerryJiehanWang:
This is a PR cloned directly from: https://github.com/NVIDIAGameWorks/kaolin/pull/345


kaolin/io/obj.py
```diff
@@ -16,37 +16,45 @@
 import warnings
 from collections import namedtuple
 
-from PIL import Image
 import numpy as np
 import torch
+from PIL import Image
+
+ObjMesh = namedtuple('ObjMesh',
```
Caenorst: Please don't rename the nametuple

kaolin/io/__init__.py
```diff
@@ -2,5 +2,9 @@
 from . import materials
 from . import obj
 from . import off
+<<<<<<< HEAD
```
Caenorst: merge conflicts?

kaolin/io/obj.py
Caenorst: I think the string should also be renamed

Caenorst:
some commits are not signed apparently

kaolin/io/obj.py
JerryJiehanWang: changed.

tests/python/kaolin/io/test_shapenet.py
```diff
@@ -0,0 +1,75 @@
...
+
+import os
+
+import pytest
+import torch
+
+from kaolin.io.obj import return_type
+from kaolin.io.shapenet import ShapeNet
+
+SHAPENET_PATH = 'data/ShapeNet'
```
Caenorst: I think we should assume `/data/ShapeNet`, `data/ShapeNet` is a relative path and should avoided.

JerryJiehanWang: Ah, I was trying to run some tests locally. Changed now.

Caenorst (approved):
LGTM! Thanks! :)

PR #379, opened 2021-04-19
Make automatic documentation check fail on warning

Caenorst:
Automatically build .rst files following kaolin structure, add some exceptions for files that are not directly included as modules.
make CI fail on warning for documentation building (some warnings were skipping the build of the documentation of whole files)

This should help keep track of any issue in the documentation build of PRs.

Remarks:
there are still some caveats, for instance, I have to exclude files in https://github.com/Caenorst/kaolin/blob/cfujitsang/auto_doc/docs/kaolin_ext.py#L21 if the module itself is not imported in the `__init__.py` for instance: https://github.com/Caenorst/kaolin/blob/cfujitsang/auto_doc/kaolin/ops/conversions/__init__.py#L1-L4 instead of:
```
from . import sdd
...
```

So we still recommend reviewers and contributors to verify documentation when adding new module files.


docs/conf.py
```diff
@@ -12,7 +12,9 @@
 #
 import os
 import sys
-sys.path.insert(0, os.path.abspath('../kaolin'))
+
+# sys.path.insert(0, os.path.abspath('../kaolin'))
```
shumash:
```suggestion
```

docs/kaolin_ext.py
```diff
@@ -0,0 +1,57 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+KAOLIN_ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), os.pardir)
+
+def run_apidoc(_):
```
shumash, line 19: I don't fully understand this file. Please add comments for developers.

kaolin/visualize/timelapse.py
```diff
@@ -14,6 +14,10 @@
 
 logger = logging.getLogger(__name__)
 
+__all__ = [
```
shumash, line 17: What is this and why is this needed? 

Please update the `doc/README.md` to reflect any tricks on excluded files and these `__all__` tricks. The current README is very nice, but is missing these details.

shumash (commented):
A few small things. 

For new modules / functions, there should be clear guidance how to update the doc.

kaolin/visualize/timelapse.py
Caenorst: This `__all__` is not only for documentation.
This affect the behavior of `from mymodule import *` that I added in https://github.com/Caenorst/kaolin/blob/cfujitsang/auto_doc/kaolin/visualize/__init__.py#L1

If you don't make an `__all__` then `from mymodule *` will return **any** variable instantiated in `mymodule`, so for instance all the imported modules: https://github.com/Caenorst/kaolin/blob/cfujitsang/auto_doc/kaolin/visualize/timelapse.py#L1-L15.

I believe this make contribution easier, to add a new feature in the file, as all the modifications will be contained in the same file (here `kaolin/visualize/timelapse.py`), instead of having to modify both the file and `__init__.py`

docs/kaolin_ext.py
Caenorst: Added some comments

kaolin/visualize/timelapse.py
Caenorst: updated docs/README.md

shumash (approved):
LGTM. 

The EXCLUDE_PATHS is a bit brittle and requires careful crafting. Please add a TODO to fix that.

Issue #389, opened 2021-04-28
PyTorch max version


mjd3:
Is there a reason for the maximum PyTorch version requirement? I've been able to install and use with PyTorch 1.8.0/1.8.1 with no issues and the max version is only listed in the actual `setup.py` file (not in the documentation online). Happy to submit a PR to update to 1.8.1 (currently 1.7.1) or remove the max version requirement entirely if that is needed! There also appears to be an environment variable "IGNORE_VER_ERR" mentioned in the comments of `setup.py` that is not used - maybe implementing this could be an option? As a side note, #344 also references this, but seems like it should be resolved by #353. 
Thanks for the awesome library!

Caenorst:
Hi @mjd3 ! Thank you for your interest in Kaolin and the kind words!

We usually restrict pytorch version to avoid having pytorch update getting us by surprise (it happened in the past). If we were to use this `IGNORE_VER_ERR` environment variable that would probably be more elegant for us to start testing new version without modifying the code.

A PR would be appreciated!

PR #390, opened 2021-04-28
Implement IGN_TORCH_VER env var in setup.py

mjd3:
Allows for ignoring the installed pytorch version when installing so that recent versions may be used while maintaining explicit version support. Addresses #389.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,8 +3,8 @@
 # TORCH_CUDA_ARCH_LIST
 #   specify which CUDA architectures to build for
 #
-# IGNORE_VER_ERR
-#   ignore version error for torch
+# IGN_TORCH_VER
+#   ignore version of PyTorch
 
 from os import environ
 from setuptools import setup, find_packages, dist
@@ -16,6 +16,7 @@ TORCH_MIN_VER = '1.5.0'
 TORCH_MAX_VER = '1.7.1'
 CYTHON_MIN_VER = '0.29.20'
 INCLUDE_EXPERIMENTAL = environ.get('KAOLIN_INSTALL_EXPERIMENTAL') is not None
+IGN_TORCH_VER = environ.get('IGN_TORCH_VER') is not None
 
 missing_modules = []
 torch_spec = importlib.util.find_spec("torch")
@@ -26,9 +27,10 @@ if torch_spec is None:
 else:
     import torch
     torch_ver = parse_version(torch.__version__)
-    if torch_ver <= parse_version(TORCH_MIN_VER) and \
-       torch_ver <= parse_version(TORCH_MAX_VER):
-        warnings.warn('Kaolin is compatible with PyTorch >= 1.5.0, '
+    if (torch_ver < parse_version(TORCH_MIN_VER) or
+       torch_ver > parse_version(TORCH_MAX_VER)) and \
+       not IGN_TORCH_VER:
+        warnings.warn(f'Kaolin is compatible with PyTorch >={TORCH_MIN_VER}, <={TORCH_MAX_VER}, '
                       f'but found version {torch.__version__} instead. '
                       'This will try to install torch in the right version. '
                       'If the installation fails we recommend to first install it.')
@@ -114,6 +116,8 @@ def get_requirements():
     if os.name != 'nt':  # no pypi torch for windows
         if os.getenv('PYTORCH_VERSION'):
             requirements.append('torch==%s' % os.getenv('PYTORCH_VERSION'))
+        elif IGN_TORCH_VER:
+            requirements.append('torch')
         else:
             requirements.append(f'torch>={TORCH_MIN_VER},<={TORCH_MAX_VER}')
     requirements.append('scipy>=1.2.0,<=1.5.2')
--- a/setup.py
+++ b/setup.py
@@ -31,12 +31,16 @@ else:
     import torch
     torch_ver = parse_version(torch.__version__)
     if (torch_ver < parse_version(TORCH_MIN_VER) or
-       torch_ver > parse_version(TORCH_MAX_VER)) and \
-       not IGNORE_TORCH_VER:
-        warnings.warn(f'Kaolin is compatible with PyTorch >={TORCH_MIN_VER}, <={TORCH_MAX_VER}, '
-                      f'but found version {torch.__version__} instead. '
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -3,8 +3,8 @@
 # TORCH_CUDA_ARCH_LIST
 #   specify which CUDA architectures to build for
 #
-# IGNORE_VER_ERR
-#   ignore version error for torch
+# IGN_TORCH_VER
```
Caenorst: I think it's better to fully spell `IGNORE` for readability

setup.py
```diff
@@ -3,8 +3,8 @@
...
+#   ignore version of PyTorch
```
Caenorst: ignore version **requirements** for pytorch

setup.py
```diff
@@ -114,6 +116,8 @@ def get_requirements():
     if os.name != 'nt':  # no pypi torch for windows
         if os.getenv('PYTORCH_VERSION'):
             requirements.append('torch==%s' % os.getenv('PYTORCH_VERSION'))
+        elif IGN_TORCH_VER:
```
Caenorst: Can you also add the same behavior at https://github.com/NVIDIAGameWorks/kaolin/blob/master/setup.py#L25 ? (with `if os.getenv('PYTORCH_VERSION')` and `if ('IGNORE_TORCH_VER')`

The problem is that if you rely on `setup.py` installing torch then it would still start by installing it between `TORCH_MIN_VER` and `TORCH_MAX_VER` (and `PYTORCH_VERSION` here doesn't seems to work neither but that's not coming from your PR).

setup.py
mjd3: Fixed.

setup.py
mjd3: Fixed.

setup.py
mjd3: Added this; let me know if you also want to remove the PYTORCH_VERSION part!

setup.py
```diff
@@ -16,19 +16,24 @@
...
 missing_modules = []
 torch_spec = importlib.util.find_spec("torch")
 if torch_spec is None:
     warnings.warn("Couldn't find torch installed, so this will try to install it. "
                   "If the installation fails we recommend to first install it.")
-    missing_modules.append(f'torch>={TORCH_MIN_VER},<={TORCH_MAX_VER}')
+    if IGNORE_TORCH_VER:
+        missing_modules.append('torch')
+    else:
+        missing_modules.append(f'torch>={TORCH_MIN_VER},<={TORCH_MAX_VER}')
```
Caenorst, line 29: If I understand this well, this will also ignore `PYTORCH_VERSION`.

Let's say in the future I want to test 1.8 and not 1.9 (assume 1.9 exists for the example), and I want to let `setup.py` installs it, that's the purpose of `PYTORCH_VERSION`, here I believe it won't work because you can't set `PYTORCH_VERSION == 1.8` without `IGNORE_TORCH_VER` and as soon as you set `IGNORE_TORCH_VER` it will try to install the latest version of torch (in this example 1.9), instead of 1.8.

I think the behavior we want is something like that:
1) `PYTORCH_VERSION` set between `TORCH_MIN_VER` and `TORCH_MAX_VER`: force (re)installation of pytorch at `PYTORCH_VERSION`
2) `PYTORCH_VERSION` set outside the supported range and `IGNORE_TORCH_VER` not set: Throw an error
3) `PYTORCH_VERSION` set outside the supported range and `IGNORE_TORCH_VER` set: Throw a gentle warning and force (re)installation of pytorch at `PYTORCH_VERSION`

PYTORCH_VERSION not set:
4) `IGNORE_TORCH_VER` not set, pytorch not installed: install pytorch at latest it can between `TORCH_MIN_VER` and `TORCH_MAX_VER`
5) `IGNORE_TORCH_VER` set, pytorch not installed: install pytorch at latest available
6) `IGNORE_TORCH_VER` not set, pytorch installed: check version of installed pytorch, if not supported: try to force reinstallation of pytorch at supported version with a gentle warning (current behavior: https://github.com/NVIDIAGameWorks/kaolin/blob/master/setup.py#L31)
7) `IGNORE_TORCH_VER` set, pytorch installed: check version of installed pytorch, if not supported: throw a gentle warning

I'm still thinking about a clean way to implement that.

Also let me Know if you need some help on testing.

Caenorst (commented):
That's a great feature, I think it still have some "bugs", see my comment. 

Since we force pytorch to be installed before running the `setup` function, I think https://github.com/NVIDIAGameWorks/kaolin/blob/master/setup.py#L114-L118 is more of sanity check.

Let me know if you need some help on testing.

setup.py
mjd3: I can definitely implement this, although I wonder about the usefulness of the PYTORCH_VERSION env variable. It seems that if you want to force an install of a specific version of PyTorch and you know exactly which version ahead of time, you should install separately (e.g., `pip install torch==PYTORCH_VERSION` followed by installing kaolin). What do you think about removing that variable entirely? Removing it would greatly simplify the logic involved and remove some redundant cases (e.g., cases where there is a current version installed and a different version specified by PYTORCH_VERSION). Let me know what you decide.

Caenorst: I see your point, it also forces a divergence of the installation process with Windows, so overall I agree.

We still need to do a hard check for pytorch being installed at the very beginning of setup.py. And we would need to fix the documentation accordingly/

mjd3: Sorry, got a bit busy with some other stuff. Pushed another commit that removes PYTORCH_VERSION entirely, add warnings for the cases described above, and adds documentation for the IGNORE_TORCH_VER env variable. Let me know if there's anything else I can do!

mjd3: @Caenorst just double-checking on this; would be great to integrate it!

Caenorst:
Hey @mjd3 I pushed a recent fix for the documentation check bug, rebase on master should solve the issue.

mjd3:
@Caenorst I rebased although it seems like the build failed (looks like it's hanging for some reason). Let me know if there's anything I can do.

Caenorst:
CI have some networking issues, I'm testing your PR on my side with all the installation cases we talked about, will approve once I'm done!

Caenorst (approved):
LGTM! Thanks for the contribution!

PR #391, opened 2021-04-29
Ignore some errors on ShapeNet

Caenorst:
ShapeNet has some errors while loading materials because of bad path / missing texture file etc... using `skip_error_handler` would spam too many warnings, so implemented a new error handler that just ignores those errors.

Fix an issue on material_order output by `obj.import_mesh`


JerryJiehanWang (approved):

JerryJiehanWang:
LGTM!

PR #397, opened 2021-05-27
Cloop/spc

charlesloop:
Add structured point clouds to Kaolin


examples/tutorial/spc_raytrace_main.py
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env python
```
Caenorst: This seems to be specific to your system, please remove

examples/tutorial/spc_raytrace_main.py
```diff
@@ -0,0 +1,131 @@
...
+
+# Copyright (c) 2019-2020, NVIDIA CORPORATION. All rights reserved.
```
Caenorst: 2019-2021 now

examples/tutorial/spc_raytrace_main.py
```diff
@@ -0,0 +1,131 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import argparse
+import torch
+import numpy as np
+import math
+import cv2
```
Caenorst: if we can, avoid opencv which is not a dependency, you can use Pillow which should have the same features (basically displaying an image right?)

Caenorst: You can also do jupyter notebook for tutorials, you can then display images inline using matplotlib.

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,399 @@
...
+from torch import nn
+from torch.autograd import Function
+import torch
+import numpy as np
+
+from kaolin import _C
+
+
+
+def ScanOctrees(octrees, olens):
```
Caenorst: We are following PEP8 conventions: https://www.python.org/dev/peps/pep-0008/#naming-conventions

python functions and variables should be of format: lowercase_with_underscores

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,399 @@
...
+    """
+    Scan batch of octree tensor.
```
Caenorst: This need a little bit more details about what it is doing (notion of "scan" may not be clear to users).

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,399 @@
...
+
+    Args:
+        octrees (torch.Tensor).dtype(torch::kByte): Shape = (b, k)
+        olens (torch.Tensor).dtype(torch::kInt32): Shape = (b)
```
Caenorst: This may need more description. Format should be:
```
name (type): some description, of shape :math:`(X, Y, Z)`
```

Usually the dtype can be expressed as a type , for instance `torch.ByteTensor`.

kaolin/csrc/utils.h
```diff
@@ -26,6 +26,30 @@
     CHECK_EQ(error, cudaSuccess) << " " << cudaGetErrorString(error); \
   } while (0)
 
+
+#define CUDA_CHECK_ARGS(condition, ...)                                        \
```
Caenorst: This looks the same than `CUDA_CHECK` above

kaolin/ops/conversions/trianglemesh.py
```diff
@@ -89,3 +91,69 @@ def trianglemeshes_to_voxelgrids(vertices, faces, resolution, origin=None, scale
...
+        >>> vertices = torch.tensor([[[0, 0, 1],
+        ...                           [0, 1, 0], 
+        ...                           [1, 0, 0]],
+        ...                
+        ...                          [[1, 1, 0], 
+        ...                           [1, 0, 1], 
+        ...                           [0, 1, 1]]], dtype=torch.float).to("cuda")
+        >>> faces = torch.tensor([[0, 1, 2]], dtype=torch.long).to("cuda")
+        >>> origin = torch.zeros((1, 3)).to("cuda")
+        >>> scale = torch.ones((1)).to("cuda")
```
Caenorst: `torch.tensor`, `torch.zeros`. `torch.ones` all have `device` argument, using `.to('cuda')` in this context imo is bad practice.

kaolin/ops/conversions/trianglemesh.py
```diff
@@ -89,3 +91,69 @@ def trianglemeshes_to_voxelgrids(vertices, faces, resolution, origin=None, scale
...
+        faces (torch.tensor): unbatched faces of shape (F, 3) of mesh to convert.
+        level (int): desired level of generated structured point cloud.
+        origin (torch.tensor): origin of the voxelgrid in the mesh coordinates. It has shape :math:`(\text{batch_size}, 3)`.
+                               Default: origin = torch.min(vertices, dim=1)[0]
+        scale (torch.tensor): the scale by which we divide the vertex position. It has shape :math:`(\text{batch_size})`.
+                              Default: scale = torch.max(torch.max(vertices, dim=1)[0] - origin, dim=1)[0]
+
+    Returns:
+        (torch.Tensor): 
+            Batched (packed) structured point clouds of shape (B, K), where k is data dependent.
```
Caenorst: this doesn't appear to be true w.r.t the code, the output seems to be of type `(torch.FloatTensor, torch.IntTensor)` It also seems that mesh_to_spc is only compatible with inputs `type(vertices) == torch.FloatTensor`  and `type(face) == torch.LongTensor`

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,399 @@
...
+
+    Returns:
+        (torch.Tensor).dtype(torch::kInt32):
+            Tensor of corresponding point lengths. Shape = (b)
+        (torch.Tensor).dtype(torch::kInt32):
+            Result of the exclusive sum of octree bytes bit counts, per batch. Shape = (b, k+b)
```
Caenorst: the output type should be `(torch.Tensor, torch.Tensor)`

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,399 @@
...
+    """    
+    return _C.ops.spc.ScanOctrees(octrees, olens)
+
+def DecodePoints(octrees, olens, plens, exsum):
+    return _C.ops.spc.DecodePoints(octrees, olens, plens, exsum)
```
Caenorst: This need some documentation, also need to follow format convention (see comment for `ScanOctrees`)

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,399 @@
...
+
+
+#  ####### Convolution ######################################
+
+
+class Conv3dFunction(Function):
+    @staticmethod
+    def forward(ctx, octrees, points, pyramid, exsum, inputs, params, kernel_vectors, jump):
+        if not inputs.is_contiguous():
+            inputs = inputs.contiguous()
```
Caenorst: you can directly do `inputs = inputs.contiguous()` AFAIK it doesn't do anything if the tensor is already contiguous, here is a test on my side:
```
>>> a=torch.rand((3, 3))
>>> b=a.contiguous()
>>> a
tensor([[0.6886, 0.4449, 0.5546],
        [0.8130, 0.2281, 0.5906],
        [0.7521, 0.9925, 0.5549]])
>>> b[:] = 0.
>>> a
tensor([[0., 0., 0.],
        [0., 0., 0.],
        [0., 0., 0.]])
```

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,399 @@
...
+        _C.ops.spc.Conv3d_forward(octrees, points, pyramid, exsum, inputs, outputs, params, kernel_vectors, jump)
+
+        ctx.octrees = octrees
+        ctx.points = points
+        ctx.pyramid = pyramid
+        ctx.exsum = exsum
+        ctx.inputs = inputs
+        ctx.params = params
+        ctx.kernel_vectors = kernel_vectors
+        ctx.jump = jump
```
Caenorst: I recommend using `ctx.save_for_backward(...)` for saving tensors, it seems to have some computational advantages.

see: https://pytorch.org/tutorials/beginner/examples_autograd/two_layer_net_custom_function.html

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,399 @@
...
+
+        return outputs
+
+    @staticmethod
+    def backward(ctx, grad_outputs):
+        if not grad_outputs.is_contiguous():
+            grad_outputs = grad_outputs.contiguous()
+
+        d_inputs = grad_outputs.new()
+        d_params = grad_outputs.new()  # this seems wrong!
```
Caenorst: What are you trying to do here? It DOES seems wrong! :p

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,399 @@
...
+
+        _C.ops.spc.Conv3d_backward(ctx.octrees, ctx.points, ctx.pyramid, ctx.exsum, ctx.inputs, d_inputs,
+                                 grad_outputs,
+                                 ctx.params, d_params,
+                                 ctx.kernel_vectors, ctx.jump )
+
+        return None, None, None, None, d_inputs, d_params, None, None
+
+
+class Conv3d(nn.Module):
```
Caenorst: I need some documentation here, especially the constructors arguments

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,399 @@
...
+    def __init__(self, in_channels, out_channels, kernel_vectors, jump=0, bias=True):
+        super(Conv3d, self).__init__()
+        self.in_channels = in_channels
+        self.out_channels = out_channels
+        self.Ksize = kernel_vectors.shape[0]
+
+        self.register_buffer('kernel_vectors', kernel_vectors)
+        self.jump = torch.tensor([jump], dtype=torch.int32)
```
Caenorst: Do you need to have `jump` as a `torch.IntTensor`? Can't it stay a python integer? (pybind11 support explicit conversion between python and C++ integer).

kaolin/ops/spc/spc.py
Caenorst: Can you add a test case or an example of usage of the Conv3d?

Caenorst (commented):
Focusing on Python side for the moment, let's focus on Conv3d because I think some comments might be redundant.

examples/tutorial/spc_raytrace_main.py
charlesloop: Removed raytracing

examples/tutorial/spc_raytrace_main.py
charlesloop: removed this raytracing file

examples/tutorial/spc_raytrace_main.py
charlesloop: removed ray tracing

kaolin/ops/spc/spc.py
charlesloop: Renamed to scan_octrees. However, I left in the name ScanOctrees in bindings.cpp. I did the same for all public facing API calls in spc.py.

kaolin/ops/spc/spc.py
charlesloop: Documentation still in progress

kaolin/ops/spc/spc.py
charlesloop: fixed

kaolin/ops/spc/spc.py
charlesloop: fixed

kaolin/ops/spc/spc.py
charlesloop: Documentation in progress

kaolin/ops/spc/spc.py
charlesloop: I fixed this. I think the reason for this was that I couldn't store an `int` in ctx.save_for_backward().

kaolin/ops/spc/spc.py
charlesloop: I added some meager test code in  test/pythoin/kaolin/spc. Needs improvement.

kaolin/ops/spc/spc.py
Caenorst: I think the string part as the python name declaration should follow python style convention. I'm ok with Cpp name since we don't have a strict style, for ScanOctrees something like:
```
ops_spc.def("scan_octrees", &ScanOctrees);
```

kaolin/csrc/ops/conversions/mesh_to_spc/mesh_to_spc_cuda.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#include <stdlib.h>
+#include <stdio.h>
+#include <torch/extension.h>
+#include <cuda_runtime.h>
```
Caenorst: This is preventing me to build Kaolin on cpu-only config

docs/modules/kaolin.ops.reduction.rst
```diff
@@ -0,0 +1,12 @@
+.. _kaolin.ops.reduction:
```
Caenorst: you can remove this file, it should be generated automatically.

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,151 @@
+.. _kaolin.ops.spc:
+
+kaolin.ops.spc
+==============
+
+.. _spc:
+
+Structured Point Clouds
```
Caenorst, line 8: Great job on the doc!

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,151 @@
...
+coordinates of octree nodes. Following the breath-first layout of the octree, the points 
+of :math:`P` are enumerated by levels in Morton order (hyperlink?). 
+Levels 0 through 8 of a structured point cloud teapot model are illustrated below:
+
+.. image:: ../img/spcTeapot.png
+
+
+A batch of octree tensors are represented by :math:`O` as a :math:`k \times 1` tensor of bytes,
+together with a :math:`b \times 1` tensor :math:`lengths`. Since :math:`O` uses packed batching,
+:math:`k` is equal to the sum the elements of :math:`lengths`.
```
Caenorst: I think there is a typo: "k is equal to the sum of the elements in lengths" (you can also directly write `sum(k)` or :math:`\sum_{i}k_i`, which may be easier to understand.

Also you can create link to references, like for packed batching, can do something like
```
ref:`packed batching<_packed>`
```
Which will generate a hyperlink to https://github.com/NVIDIAGameWorks/kaolin/blob/master/docs/modules/kaolin.ops.batch.rst#packed (see to understand the reference), with the text "packed batching".

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,151 @@
...
+performace. Next, these tensors are input to
+
+>>> points = spc.decode_points(octree, pyramid, exsum)
+
+The tensor :attr:`points` contains the batches of point hierachies.
+
+Structured point cloud data can be derived from multiple sources. Using mesh_to_spc() will
+convert a triangle mesh model into octree tensor :math:`O`. We can also sonstruct :math:`O`
+from unstructured point cloud data (coming soon?). Or from an implicit function 
+(need a cuda call back).
```
Caenorst: I think it could be interesting to show a simple example to show the usage of those auxiliary tensors

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,151 @@
...
+kaolin.ops.spc
+==============
+
+.. _spc:
+
+Structured Point Clouds
+-----------------------
+
+Structured Point Clouds are a tensor representation of a point cloud hierarchy 
+:math:`P` defined by a sparse octree tensor :math:`O`.
```
Caenorst: Is there a direct relation between `O` / `P` and the real variables that we are using? Shouldn't we use those real variables names instead?

I find the transition between the two notations a little bit abrupt.

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,151 @@
...
+
+Structured point cloud data can be derived from multiple sources. Using mesh_to_spc() will
+convert a triangle mesh model into octree tensor :math:`O`. We can also sonstruct :math:`O`
+from unstructured point cloud data (coming soon?). Or from an implicit function 
+(need a cuda call back). 
+
+Feature Data
+............
+
+We associate feature data with levels in the point hierarchy. This might be a tensor
```
Caenorst: That's a little bit too verbose, needs an example I think.

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,151 @@
+.. _kaolin.ops.spc:
+
+kaolin.ops.spc
```
Caenorst, line 3: Some general tips:
- You can make a list using the `*` character
- You can reference specific part of the whole documentation with `:ref:`, an attribute / variable with `:attr:`, a function with `:func:`, that can create an hyperlink but also slightly change the font
- You can make the font **bold** with `**` or *italic* with `*`
- When naming a variable or some code you can use `` to highlight it.

You can look at the way I used those in: https://raw.githubusercontent.com/NVIDIAGameWorks/kaolin/master/docs/modules/kaolin.ops.batch.rst and the resulting generated doc in: https://kaolin.readthedocs.io/en/latest/modules/kaolin.ops.batch.html

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,151 @@
...
+structured point clouds. 
+
+
+Convolutions
+............
+
+We provide several sparse convolution layers for structured point clouds.
+Convolutions are characterized by the size of the input and output channels, 
+an array of kernel vectors, and level jump. 
+An example of how to create a :math:`3 \times 3 \times 3` kernel follows.
```
Caenorst: this would benefit of highlighting variables using `` or bold font.

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,151 @@
...
+
+>>> Kvec = []
```
Caenorst: I'm struggling to understand the relation between `K`, `Kvec` and `kernel_vectors`

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,151 @@
...
+tensor([[-1, -1, -1],
+        [-1, -1,  0],
+        [-1, -1,  1],
+        ...
+        ...
+        [ 1,  1, -1],
+        [ 1,  1,  0],
+        [ 1,  1,  1]], device='cuda:0', dtype=torch.int16)
+
+The kernel vectors determine the shape of the convolution kernel.
```
Caenorst, line 189: What do you mean by: "The kernel vectors determine the shape of the convolution kernel."

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,151 @@
...
+
+.. math::
+
+  n^T(i,k) = \text{ID}\left(P_i-\overrightarrow{K}_k\right)
+
+
+The value :attr:`jump` is used to indicate the difference in levels between the iput features
+and the output features. For convolutions, this is the number of levels to downsample; while
+for transposed convolutions, :attr:`jump` is the number of levels to upsample. The value of :attr:`jump` must
+be positive, and may not go beyond the highest level of the octree.
```
Caenorst, line 213: I think the whole part of `Convolution` could be in the description of the Ops.

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+import torch
+import numpy as np
+
+from kaolin import _C
+
+__all__ = ['scan_octrees', 'decode_points', 'conv', 'conv_transpose', 'max_pool', 'max_unpool', 'avg_pool', 'to_dense']
+
+def scan_octrees(octrees, olens):
+    r"""Scan batch of octrees tensor. There are two steps. First, a list is formed 
+    containing the number of set bits in each octree btye. Second, the exclusive
```
Caenorst: typo: b**yt**e

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+from torch.autograd import Function
+import torch
+import numpy as np
+
+from kaolin import _C
+
+__all__ = ['scan_octrees', 'decode_points', 'conv', 'conv_transpose', 'max_pool', 'max_unpool', 'avg_pool', 'to_dense']
+
+def scan_octrees(octrees, olens):
+    r"""Scan batch of octrees tensor. There are two steps. First, a list is formed
```
Caenorst: Usually we separate the first sentence (coarse definition) from the paragraph (detailed description) with a line break.

kaolin/ops/spc/spc.py
Caenorst: I believe this is still wrong, the output type should be `(torch.Tensor, torch.Tensor)` or `(torch.IntTensor, torch.IntTensor)`, not just `(torch.Tensor)`

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+        olens (torch.Tensor).dtype(torch::kInt32): a list of the lengths of each inididual octree in input batchShape = (b)
+
+    Returns:
+        (torch.Tensor): 
+            A batch of tensor containing the level sizes and offsets,
+            and a tensor of batch exclusive sums of octree bit counts.
+    
+    .. note::
+
+        The returned tensor of exclusive sums is padded with an extra elment for each
```
Caenorst: typo: el**e**ment

I'm not sure to understand how this is padded but the example should show that easily.

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+        item in the batch.  
+
+    Example:
+        >>> teapot = np.load("teapot.npz")
+        >>> octree = torch.from_numpy(teapot["octree"], device=torch.device('cuda:0'))
+        >>> lengths = torch.tensor([len(octree)], dtype=torch.int32)
```
Caenorst: This example section is not showing `scan_octrees`.

My suggestion for example section is that it should show:
1) How to use the function
2) How is it behaving

Typically, if possible, it should show a hard-coded input tensor and show the output of the function, so user can infer the behavior from the `inputs` => `outputs` relation. (For instance I can see the padding that you are referring to in the note).

See: https://github.com/charlesloop/kaolin/blob/cloop/spc/kaolin/ops/batch.py#L52-L58 for example

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+        >>> octree = torch.from_numpy(teapot["octree"], device=torch.device('cuda:0'))
+        >>> lengths = torch.tensor([len(octree)], dtype=torch.int32)   
+
+    """
+    return _C.ops.spc.ScanOctrees(octrees, olens)
+
+
+def decode_points(octrees, pyramid, exsum):
+    r"""Decode batched octree into batch of structured point hierarchies,
+    and batch of book keepinh pyramids.
```
Caenorst: typo: keepin**g**

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+__all__ = ['scan_octrees', 'decode_points', 'conv', 'conv_transpose', 'max_pool', 'max_unpool', 'avg_pool', 'to_dense']
+
+def scan_octrees(octrees, olens):
+    r"""Scan batch of octrees tensor. There are two steps. First, a list is formed 
+    containing the number of set bits in each octree btye. Second, the exclusive 
+    sum of this list is taken. The output of this function is used in 
+    :attr:`DecodePoints`.
+
+    Args:
+        octrees (torch.Tensor).dtype(torch::kByte): batched (packed) octrees of shape = (k), where k is number of bytes in octrees.
```
Caenorst: type should be `(torch.ByteTensor)`

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+        olens (torch.Tensor).dtype(torch::kInt32): a list of the lengths of each inididual octree in input batchShape = (b)
```
Caenorst: type should be `(torch.IntTensor)`

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+
+
+def decode_points(octrees, pyramid, exsum):
+    r"""Decode batched octree into batch of structured point hierarchies,
+    and batch of book keepinh pyramids.
+
+    Args:
+        octrees (torch.Tensor).dtype(torch::kByte): Shape = (k)
+        pyramid (torch.Tensor).dtype(torch::kInt32): Shape = (b, 2, l+2)
+        exsum (torch.Tensor).dtype(torch::kInt32): Shape = (k+b)
```
Caenorst: syntax for type is not correct.

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+
+    Returns:
+        (torch.Tensor);
+            A tensor containing a batched point hierachy derived from a bacth of octrees
```
Caenorst: typos:
hiera**r**chy
ba**tc**h

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+
+        d_inputs, d_params = _C.ops.spc.Conv3d_backward(octrees, points, pyramid, exsum, inputs,
+                                 grad_outputs,
+                                 params,
+                                 kernel_vectors, ctx.jump )
+
+        return None, None, None, None, d_inputs, d_params, None, None
+
+
+class conv(nn.Module):
```
Caenorst: following PEP8 (https://www.python.org/dev/peps/pep-0008/#naming-conventions): 
classes: CamelCase
functions and variables: lower_case_with_underscore

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+    where :math:`w_k` are weights associated with the kernel.
+
+    Args:
+        in_channels (int): the number of input channels in the
+            input tensor.
+
+        out_channels (int): the number of output channels in the
+            output tensor.
+
+        kernel_vectors (short): a python list of 3D offsets that
```
Caenorst: There is a conflict between the type and the description / code.

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+            input (`torch.FloatTensor`): 
+
+        Returns:
+            (torch.Tensor): 
+                Batched (packed) output of convolution.
+
+        Example:
+            >>> # Kernel vector tensor K defined as above
+            >>> m = spc.conv(3, 16, K, jump=1)
+            ...
```
Caenorst: not necessary

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+            >>> # octrees, points, pyramid, exsum as above
+            >>> level = pyramid.size(2)-2
+            >>> size = pyramid[:,0,level].sum().item()
+            >>> input = torch.randn((size, 3, device=torch.device('cuda:0'))
+            >>> output = m(input)
```
Caenorst: It would be great to display values of `input`, and the kernel inside the convolution layer.

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+            octrees (`torch.ByteTensor`): structured point cloud batched octrees
+
+            points (`torch.ByteTensor`): structured point cloud batched points
+
+            pyramid (`torch.IntTensor`): structured point cloud batched pyarmids
+
+            exsum (`torch.IntTensor`): structured point cloud batched octree 
+                bit count exclusive sum
+
+            input (`torch.FloatTensor`):
```
Caenorst: Missing some description, "`" character is not necessary for the type.

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+        return self.__class__.__name__ + s
+
+
+#  ####### Transposed Convolution ############################
+
+
+class ConvTranspose3dFunction(Function):
+    @staticmethod
+    def forward(ctx, octrees, points, pyramid, exsum, inputs, params, kernel_vectors, jump):
+        if not inputs.is_contiguous():
```
Caenorst: don't need this `if`

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+            inputs = inputs.contiguous()
+
+        ctx.save_for_backward(octrees, points, pyramid, exsum, inputs, params, kernel_vectors)
+        ctx.jump = jump
+
+        return _C.ops.spc.ConvTranspose3d_forward(octrees, points, pyramid, exsum, inputs, params, kernel_vectors, jump)
+
+    @staticmethod
+    def backward(ctx, grad_outputs):
+        if not grad_outputs.is_contiguous():
```
Caenorst: don't need this `if`

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,631 @@
...
+            output tensor.
+
+        kernel_vectors (short): a python list of 3D offsets that 
+            define the shape of the kernel.
+
+        jump (int, optional): the difference between the input and
+            output levels for the convolution
+
+        bias (bool, optional): if True, the convolution layer
+            has a bias.
```
Caenorst: Add: "Default: True"

kaolin/csrc/ops/conversions/mesh_to_spc/mesh_to_spc_cuda.cpp
charlesloop: offending header file include removed

docs/modules/kaolin.ops.reduction.rst
charlesloop: Not sure what this file has to do with my stuff. No action taken

docs/modules/kaolin.ops.spc.rst
charlesloop: Thanks!

docs/modules/kaolin.ops.spc.rst
charlesloop: Done

docs/modules/kaolin.ops.spc.rst
charlesloop: Better now?

docs/modules/kaolin.ops.spc.rst
charlesloop: I think it's better now

kaolin/ops/spc/spc.py
charlesloop: Better, maybe

docs/modules/kaolin.ops.reduction.rst
Caenorst: I think you added it by mistake, you can do `git rm docs/modules/kaolin.ops.reduction.rst` to remove it from the MR.

kaolin/csrc/bindings.cpp
```diff
@@ -35,6 +37,22 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
...
+    ops_spc.def("ConvTranspose3d_forward", &ConvTranspose3d_forward);
+    ops_spc.def("ConvTranspose3d_backward", &ConvTranspose3d_backward);
+    ops_spc.def("MaxPool3d_forward", &MaxPool3d_forward);
+    ops_spc.def("MaxPool3d_backward", &MaxPool3d_backward);
+    ops_spc.def("MaxUnpool3d_forward", &MaxUnpool3d_forward);
+    ops_spc.def("MaxUnpool3d_backward", &MaxUnpool3d_backward);
+    ops_spc.def("AvgPool3d_forward", &AvgPool3d_forward);
+    ops_spc.def("AvgPool3d_backward", &AvgPool3d_backward);
+    ops_spc.def("ToDense_forward", &ToDense_forward);
+    ops_spc.def("ToDense_backward", &ToDense_backward);
```
Caenorst: python binding declaration should follow PEP8: lower_case_with_underscore. You can keep the same convention for the C++ part. I would really prefer avoid mixing CamelCase with lower_case_with_underscore in a name.

example: `ops_spc.def('to_dense_backward", &ToDenseBackward);`

kaolin/csrc/ops/conversions/mesh_to_spc/mesh_to_spc.h
```diff
@@ -0,0 +1,29 @@
...
+#define KAOLIN_OPS_CONVERSIONS_MESH_TO_SPC_H_
+
+#include <ATen/ATen.h>
+
+namespace kaolin {
+
+std::vector<at::Tensor>  mesh_to_spc(
+    at::Tensor vertices,
+    at::Tensor triangles,
+    uint Level) ;
```
Caenorst, line 30: argument name shouldn't start with a capital case.

kaolin/csrc/ops/conversions/mesh_to_spc/mesh_to_spc_cuda.cpp
```diff
@@ -0,0 +1,136 @@
...
+#define CHECK_TRIPLE(x) TORCH_CHECK(x.dim() == 2 && x.size(1) == 3, "input is not Nx3")
+#define CHECK_FLOAT(x) TORCH_CHECK(x.scalar_type() == at::ScalarType::Float, "input is not Float")
+#define CHECK_LONG(x) TORCH_CHECK(x.scalar_type() == at::ScalarType::Long, "input is not Long")
+#define CHECK_PACKED_FLOAT3(x) CHECK_CUDA(x); CHECK_CONTIGUOUS(x); CHECK_FLOAT(x) CHECK_TRIPLE(x)
+#define CHECK_PACKED_LONG3(x) CHECK_CUDA(x); CHECK_CONTIGUOUS(x); CHECK_LONG(x) CHECK_TRIPLE(x)
+
+using namespace std;
+using namespace torch::indexing;
+
+extern ulong GetStorageBytes(void* d_temp_storageA, morton_code* d_M0, morton_code* d_M1, uint max_total_points);
```
Caenorst: I don't think declaring it `extern` is necessary

kaolin/csrc/spc_math.h
```diff
@@ -0,0 +1,273 @@
...
+#define KAOLIN_SPC_MAX_POINTS    (0x1<<27)
+
+
+typedef unsigned char		uchar;
+typedef unsigned short		ushort;
+typedef unsigned int		uint;
+
+typedef long unsigned int	morton_code;
+typedef short3				point_data;
+typedef long3               tri_index;
```
Caenorst, line 34: something went bad with spacing here

kaolin/csrc/spc_math.h
```diff
@@ -0,0 +1,273 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#ifdef WITH_CUDA
+
+#include <vector_types.h>
+
+using namespace std;
```
Caenorst, line 21: Don't write using namespace at global scope in a header file:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf7-dont-write-using-namespace-at-global-scope-in-a-header-file

kaolin/csrc/ops/conversions/mesh_to_spc/mesh_to_spc_cuda.cpp
```diff
@@ -0,0 +1,136 @@
...
+  torch::Tensor F = torch::zeros({3*ntris}, torch::device(torch::kCUDA).dtype(torch::kFloat));
+  
+  torch::Tensor axis = torch::zeros({ntris}, torch::device(torch::kCUDA).dtype(torch::kByte));
+  torch::Tensor W = torch::zeros({ntris}, torch::device(torch::kCUDA).dtype(torch::kInt16));
+  torch::Tensor pmin = torch::zeros({2*ntris}, torch::device(torch::kCUDA).dtype(torch::kInt16));
+
+  torch::Tensor Odata = torch::zeros({KAOLIN_SPC_MAX_OCTREE}, torch::device(torch::kCUDA).dtype(torch::kByte));
+  torch::Tensor Pyramid = torch::zeros({2, Level+2}, torch::device(torch::kCPU).dtype(torch::kInt32));
+
+  // get tensor data pointers
```
Caenorst: The way I see it, for CUDA ops, the `.cpp` should only contains torch tensors and so the conversion from tensor => data_ptr should be in the `.cu`

The main reason is that when you want to support multiple types the DISPATCH macro have to be in the .cu (you can't pass a non-specialized template through symbol)

The 2nd reason here, is that you are using float3 which are CUDA specific.

kaolin/csrc/ops/conversions/mesh_to_spc/mesh_to_spc_cuda.cpp
```diff
@@ -0,0 +1,136 @@
...
+    uint Level) 
+{
+  CHECK_PACKED_FLOAT3(vertices);
+  CHECK_PACKED_LONG3(triangles);
+
+  uint npnts = vertices.size(0);
+  uint ntris = triangles.size(0);
+
+  // allocate local GPU storage
+  torch::Tensor P = torch::zeros({KAOLIN_SPC_MAX_POINTS, 3}, torch::device(torch::kCUDA).dtype(torch::kInt16));
```
Caenorst: for multi-gpu support, I would recommend using `vertices.options()` instead of `torch::device(torch::kCUDA)`

kaolin/csrc/ops/conversions/mesh_to_spc/mesh_to_spc_cuda_kernel.cu
```diff
@@ -0,0 +1,507 @@
...
+#include "../../../spc_math.h"
+
+namespace kaolin {
+#ifdef WITH_CUDA
+
+using namespace cub;
+using namespace std;
+
+
+ulong GetStorageBytes(void* d_temp_storageA, morton_code* d_M0, morton_code* d_M1, uint max_total_points)
```
Caenorst: Can you add a short comment description of all the functions / kernels in this file? Add some descriptions of the arguments too please.

kaolin/csrc/ops/spc/spc_conv.h
```diff
@@ -0,0 +1,251 @@
...
+template <typename Itype> using InOutMaps = vector<vector<Itype>>;
+
+// Input index to output index mapping in ptr, sise pair
+// Used for device pointer and size
+template <typename Itype> using pInOutMaps = vector<pVector<Itype>>;
+
+
+ulong GetStorageBytesX(void* d_temp_storage, uint* d_Info, uint* d_PrefixSum, uint max_total_points);
+
+int scan_octrees_cuda(
```
Caenorst: Please add short description with argument descriptions

tests/python/kaolin/spc/avg_pool_test.py
```diff
@@ -0,0 +1,163 @@
+from __future__ import division
```
Caenorst: test structure should follow directory tree of core library, the name of the python file have to start with `test_`

I only see the file `kaolin/ops/spc/spc.py` so it should be `tests/python/kaolin/spc/test_spc.py`

tests/python/kaolin/spc/compare.py
```diff
@@ -0,0 +1,198 @@
+import math
```
Caenorst: What is this python module doing? It's not a test file so it doesn't belong here.

tests/python/kaolin/spc/conv_test.py
```diff
@@ -0,0 +1,143 @@
+from __future__ import division
```
Caenorst: same comment than for `avg_pool_test.py`

tests/python/kaolin/spc/conv_transpose_test.py
```diff
@@ -0,0 +1,139 @@
+from __future__ import division
```
Caenorst: same comment than for `avg_pool_test.py`

tests/python/kaolin/spc/max_pool_test.py
```diff
@@ -0,0 +1,149 @@
+from __future__ import division
```
Caenorst: same comment than for `avg_pool_test.py`

examples/tutorial/spc_to_dense_test.py
```diff
@@ -0,0 +1,50 @@
+import torch
```
Caenorst: example shouldn't have the prefix "test", as an example should go beyond the scope of testing.

examples/tutorial/spc_to_mesh_test.py
```diff
@@ -0,0 +1,36 @@
+import torch
```
tovacinni: Filename typo? :)

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,178 @@
...
+
+kaolin.ops.spc
+==============
+
+.. _spc:
+
+Structured Point Clouds
+-----------------------
+
+Structured Point Clouds are an octree-base representation of a batch of point cloud hierarchy
```
shumash: It is confusing if you just bold variable names, because "batch of point cloud hierarchy points" does not read like proper english. Please use code style for variable names. Also, it is absolutely unclear what a "point cloud hierarchy" is. This first paragraph should make it clear to the reader if this representation is useful for them.

Here is what I had for the SIGGRAPH talk -- please make this even more concrete.
"This Kaolin library release additionally includes Structured Point Clouds data structure that improves efficiency. This octree-based structure can store arbitrary information, like sparse voxel grids and quantized point clouds. It already supports a number of operations, including efficient ray-tracing, with more to come. "

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,178 @@
...
+
+Structured Point Clouds are an octree-base representation of a batch of point cloud hierarchy 
+**points**
+
+.. warning::
+  Structured Point Clouds internal layout and structure is still experimental may be modified.
+
+.. _octree-text:
+
+We represent **octrees** as a :math:`k \times 1` tensor of bytes.
```
shumash: Here and everywhere, code notation for variables. 

Also, it is the first time octrees is mentioned. First say what it is and then explain how it's represented.

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,178 @@
...
+-----------------------
+
+Structured Point Clouds are an octree-base representation of a batch of point cloud hierarchy 
+**points**
+
+.. warning::
+  Structured Point Clouds internal layout and structure is still experimental may be modified.
+
+.. _octree-text:
+
```
shumash, line 30: The first section should be high-level. How it created? What does it store? What does it allow you to do? The following in-depth section is a great description that should follow this high-level bit. @Caenorst Could you add the high-level part, please?

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,178 @@
...
+Structured point cloud data can be derived from multiple sources. Using 
+:func:`kaolin.ops.conversions.trianglemeshes_to_spcs`
+will
+convert a batched triangle mesh model into a batched octree tensor **octrees**. 
+We can also construct **octrees**
+from unstructured point cloud data, or from the level set of an implicit function :math:`f(x,y,z)`. 
+
+
+Since **octrees** use :ref:`packed` batching, we need a :math:`b \times 1` tensor **lengths**
```
shumash: Please check you switch to code notation fo variables. This is *really hard* to parse.

docs/modules/kaolin.ops.spc.rst
```diff
@@ -0,0 +1,178 @@
...
+The tensor **points** contains the batches of point hierachies. The tensor **points**, **octrees**,
+**pyramid**, and **exsum** contains cached geometry and structure data needed for convolutions
+between octree levels.
+
+
+Feature Data
+............
+
+We associate feature data with levels in the point hierarchy. This might be a tensor
+of RGB color or normal data corresponding to the highest octree level. Data for
```
shumash: Wait... you can't store things in the leaves? I am confused.

docs/modules/kaolin.ops.spc.rst
shumash: Also, this part -- about what can be stored, and how it is retrieved -- belongs at the very top of documentation.

shumash (changes requested):
Reviewed top-of-the-file docs. API comments coming soon.

docs/modules/kaolin.ops.spc.rst
Caenorst: The octree structure is a way to store `coordinates`. You can select a specific level and you will have a bunch of coordinates (with a kind of "coarseness" depending on the level). Then you will associate features which are stored in a coalescent tensor (so technically the octree don't "contain" the features), the order of those features are associated to the morton order of the coordinates.

This is similar to sparse tensor where you one tensor `indices` and one tensor `values`. (The octree would be `indices`)

Very often you will start with the coordinates associated to the leaves, but when you apply convolution you can use `jump` so the output features will be associated to the coordinates at `output_level` = `input_level` - `jump`. (And then you can use `conv_transpose3d` with jump where the output features will be associated to the coordinates at `output_level` = `input_level` + `jump`.

`jump` act as a "dimension reduction" as `stride` would be in standard convolution.

I believe different level are also used for NeuralLOD where the ray-tracing can be applied at those different level-of-details.

kaolin/ops/conversions/trianglemesh.py
```diff
@@ -89,3 +92,79 @@ def trianglemeshes_to_voxelgrids(vertices, faces, resolution, origin=None, scale
...
+        ...                           [1, 0, 1],
+        ...                           [0, 1, 1]]], dtype=torch.float)
+        >>> faces = torch.tensor([[0, 1, 2]], dtype=torch.long)
+        >>> origin = torch.zeros((1, 3))
+        >>> scale = torch.ones((1))
+        >>> trianglemeshes_to_spcs(vertices, faces, level=2)
+        tensor([23, 128, 22,  22,  22, 232, 104, 104, 104, 1], device='cuda:0', dtype=torch.uint8),
+        tensor([5, 5])
+
+    .. image:: ../img/mesh_to_spc.png
```
shumash: This figure really helps. Thanks for adding!

kaolin/ops/spc/convolution.py
```diff
@@ -0,0 +1,335 @@
...
+__all__ = [
+    'conv3d',
+    'Conv3d',
+    'conv_transpose3d',
+    'ConvTranspose3d',
+]
+
+class Conv3dFunction(Function):
+    @staticmethod
+    def forward(ctx, octrees, points, level, pyramid, exsum, inputs, params, kernel_vectors, jump):
```
shumash: **I strongly recommend a container for these variables.** There are simply too many: imagine using 2 different SPCs in the training code (e.g. for train and eval). 

For example, can be really lightweight:
```
class StructuredPointCloud:
   def __init__(self, ...):
       self.octrees = ...
       self.points = ...

   def to_dict(self):
       retun {'octrees': self.octrees, 'points': self.points ...}
  # Note: or can override **, but it's more bug prone and harder to debug
```

Then, to maintain backward compatibility, you can recommend the following usage of low-level functions:
```
myspc = StructuredPointCloud(...)
conv = Conv3dFunction()

res = conv(**myspc.to_dict())
```

I believe the only change required is ignoring extra args with expected names for functions that don't need them.

kaolin/ops/spc/convolution.py
```diff
@@ -0,0 +1,335 @@
...
+        octrees, points, pyramid, exsum, inputs, params, kernel_vectors = ctx.saved_tensors
+
+        d_inputs, d_params = _C.ops.spc.Conv3d_backward(octrees, points, ctx.level, pyramid, exsum, inputs,
+                                 grad_outputs,
+                                 params,
+                                 kernel_vectors, ctx.jump )
+
+        return None, None, None, None, None, d_inputs, d_params, None, None
+
+def conv3d(octrees, points, level, pyramid, exsum, input,
```
shumash: Add docstring. Specifically, dimensions of bias should be listed.

kaolin/ops/uint8.py
```diff
@@ -0,0 +1,66 @@
...
+# limitations under the License.
+
+
+from itertools import product
+import torch
+
+global _uint8_to_bits_luts
+_uint8_to_bits_luts = {}
+
+def uint8_to_bits(uint8_t):
```
shumash: This seems quite specific to spc implementation. If this is correct, please let's not create a separate module with a generic name `uint8`, but keep these functions under spc.

kaolin/render/spc/raytrace.py
```diff
@@ -0,0 +1,47 @@
...
+
+import torch
+import torch.autograd
+from torch.autograd import Function
+
+from kaolin import _C
+
+
+def unbatched_raytrace(octree, points, pyramid, exsum, origin, direction, level):
+    r"""Apply ray tracing over an SPC structure.
```
shumash: The type of the output needs to be listed, including its shape. I.e. if anything can be stored in the SPC, what exactly does the raytrace return and how can it be used. A short example with inline comments would help.

tests/python/kaolin/spc/compare.py
shumash: Can go under `utils/testing_spc.py` or similar.

tests/python/kaolin/spc/spc_raytrace_main.py
```diff
@@ -0,0 +1,145 @@
...
+                        help='Path to spc .npz file.')
+
+
+    args = parser.parse_args()
+
+
+    F = np.load(args.filepath)
+    octree = torch.from_numpy(F["octree"]).to("cuda")
+
+    # colors = torch.from_numpy(F["colors"])
```
shumash: needs clean up

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,245 @@
...
+    .. note::
+
+        The returned tensor of exclusive sums is padded with an extra element for each
+        item in the batch.
+    """
+    return _C.ops.spc.ScanOctrees(octrees.contiguous(), lengths.contiguous())
+
+def generate_points(octrees, pyramid, exsum):
+    r"""Generate the point data for a structured point cloud.
+    Decode batched octree into batch of structured point hierarchies,
```
shumash, line 69: It is not clear what a structured point hierarchy is. Please define, also list the shape of the output and its type.

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,245 @@
...
+
+    Args:
+        octrees (torch.ByteTensor):
+            Batched :ref:`packed<packed>` collection of octrees of shape :math:`(\text{num_node})`.
+        lengths (torch.IntTensor):
+            The number of nodes per octree.
+            of shape :math:`(\text{batch_size})`.
+
+    Returns:
+        (int, torch.IntTensor, torch.IntTensor):
```
shumash, line 50: Please include shape information.

shumash (changes requested):
This is quite a substantial piece of work. Really excited to have this in the library!

I only commented on the API and the docs.

kaolin/ops/spc/convolution.py
Caenorst: Please see my draft here: https://github.com/charlesloop/kaolin/blob/cloop/spc/kaolin/rep/spc.py

Please give some feedback.

kaolin/ops/uint8.py
Caenorst: I moved the module in `kaolin/ops/spc/`. still kept it in its own file `uint8.py` but the functions are forwarded to `kaolin.ops.spc` namespace

kaolin/ops/spc/spc.py
Caenorst: Add some better definition of `points` in the introduction. Please give feedback.

tests/python/kaolin/spc/spc_raytrace_main.py
Caenorst: This file will be removed, this is not an official test.

kaolin/ops/spc/spc.py
shumash: Almost perfect, but you still need to define "structured point hierarchies". In the main doc, where you have "points: torch.IntTensor of shape  (...) correspond to the sparse coordinates at all levels," **add something like: "We refer to this packed <<link here>> tensor as structured point hierarchies."**

kaolin/ops/spc/convolution.py
shumash: Looks good. Please go ahead.

docs/modules/kaolin.ops.spc.rst
shumash: The first paragraph of this is clear and great. Can you add something like this around where you mention Morton order, please?

docs/modules/kaolin.ops.spc.rst
```diff
@@ -1,123 +1,134 @@
...
+=======
 
-We represent **octrees** as a :math:`k \times 1` tensor of bytes. 
-Each byte **octrees[i]** represents an octree node corresponding to a 
-:math:`2\times 2\times 2` partition of a cube of space. 
-Each bit indicates the binary occupancy of a child subcube as depicted below. 
+.. _spc_octree:
+
+An `octree <https://en.wikipedia.org/wiki/Octree>`_ is a tree data structure where each node have 8 childrens.
+
```
shumash: Rm paragraph break.

docs/modules/kaolin.ops.spc.rst
```diff
@@ -1,123 +1,134 @@
...
+i.e: each node represente a partitioning of its 3D space (partition) of :math:`(2, 2, 2)`.
+
+.. figure:: ../img/octree.png
+  :scale: 30 %
+  :alt: An octree 3D partitioning
+
+  Credit: https://en.wikipedia.org/wiki/Octree
+
+Currently Structured Point Clouds are used to represent 3D surfaces, and so all the leaves are at the same ``level`` (depth).
+
```
shumash, line 65: Rm paragraph break.

docs/modules/kaolin.ops.spc.rst
```diff
@@ -1,123 +1,134 @@
...
+This allow very efficient processing on GPU, with custom CUDA kernels, for ray-tracing and convolution.
+
+The structure is containing more finer details as you go deeper in the tree, below are the Levels 0 through 8 of a structured point cloud teapot model:
+
+.. image:: ../img/spcTeapot.png
+
+We represent ``octrees`` as a tensor of bytes.
```
shumash: This reads out of order. Here add the high-level part about storing coordinates and then features in the same order. Otherwise, you start with octrees without explaining what an SPC is exactly or what data components it has. 

You can also introduce the container class here.

shumash (changes requested):
A few minor things. Please address.

Thanks for making lots of good changes.

kaolin/ops/spc/convolution.py
Caenorst: Done

docs/modules/kaolin.ops.spc.rst
Caenorst: > Here add the high-level part about storing coordinates and then features in the same order.

Technically the SPC can exist without features, like binary voxelgrids or point clouds (although I would agree that in frequent usages that the SPC will be used with features associated to the coordinates, the ray tracing for instance do not necessitate features), which is why:
- I compare the SPC to sparse voxelgrids and quantized point clouds in the first paragraph
- In the `Feature Data` section I wrote "to those coordinates **may** correspond a set of features"
- The high level class do not contain any features.

So I would consider the features to be **associated** to the SPC, but that's just a preference, I can have the features as part of the SPC too. (And so add `features` to `kaolin.rep.Spc` with an associated `level`).

> Otherwise you can start with octrees

The documentation is about the SPC, the octrees being a (the most important) *component* of the SPC, I think it would be a little bit weird to talk about the octree without contextualizing it to the SPC in the documentation for `kaolin.ops.spc.rst`


Following your recommendation, I can only talk about the octree **theoretical structure** + associated features (high level) in the first section (still after the short SPC introduction).

Then in **Related attributes** I can introduce `octrees` and `features` as python attributes.

> You can also introduce the container class here.

I added a short note but I'm considering moving the whole doc to `kaolin.ops.rep`, what do you think?

Caenorst:
@shumash  I noticed something missing from my documentation, the `octrees` bytes represent the presence / location of childrens, but the number of bytes /= number of nodes, because the last bytes still have childrens (and so the last "nodes" technically don't have any bytes). Does this require explanation? Is so do you have any suggestion?

Caenorst:
https://github.com/NVIDIAGameWorks/kaolin/pull/397#discussion_r682256992 github doesn't let me answer over this comment.

I'm not sure what part you are referring to:

1)
> We represent ``octrees`` as a tensor of bytes.
Each byte ``octrees[i]`` represents an octree node where each bit indicate the binary occupancy of a child node / partition in `Morton Order <https://en.wikipedia.org/wiki/Z-order_curve>`_ as depicted below.
2)
> for each octree, the nodes / bytes are following breadth-first-search order (with Morton order for childrens order), the octree bytes are then :ref:`packed` to form ``octrees``.

I think otherwise I can add this information in the very first paragraph of octrees:
> An `octree <https://en.wikipedia.org/wiki/Octree>`_ is a tree data structure where each node have 8 childrens.
We use this structure to do a recursive three-dimensional space partitioning,
i.e: each node represente a partitioning of its 3D space (partition) of :math:`(2, 2, 2)`.
The octree then contains the information necessary to find the sparse coordinates.

kaolin/render/spc/raytrace.py
```diff
@@ -48,4 +50,43 @@ def unbatched_raytrace(octree, points, pyramid, exsum, origin, direction, level)
         direction.contiguous(),
         level)
 
+def mark_first_hit(nuggets):
+    r"""Mark the first hit in the nuggets. 
+
+    Returns:
+        first_hits (torch.BoolTensor):
```
Caenorst: @tovacinni can you add some details to the descriptions? 
- What does "Mark the first hit" means?
- Can you add some description for the arguments and return ? (shape will be useful too)

kaolin/render/spc/raytrace.py
```diff
@@ -48,4 +50,43 @@ def unbatched_raytrace(octree, points, pyramid, exsum, origin, direction, level)
...
+        level (int): level of the SPC to trace.
+        info (torch.BoolTensor): First hits. Will compute if None.
+        info_idxes (torch.IntTensor): Packed indices of first hits. Will compute if None.
+        mask (torch.BoolTensor): Mask to determine if the ray is still alive. Default to false if None.
+        init (bool): First run? Enables slight optimizations.
+
+    Returns
+        distance (torch.FloatTensor): distance from ray origin to ray-aabb intersection.
+        pidx (torch.LongTensor): corresponding point index.
+        mask (torch.BoolTensor): rewritten mask.
```
Caenorst: Can you add some shape informations to the input / outputs ?

docs/modules/kaolin.ops.spc.rst
shumash: Here's how a doc would read to a new reader:

1. great high-level paragraph, sounds useful
2. "Octrees section" I guess they are providing some background on octrees... Wait, they are also mentioning a variable `octrees`... how does this relate to SPC? Is this SPC itself?

You see what I mean? The reader has to do a lot of unnecessary piecing together, as these components are not introduced. 

**It would be sufficient to add a 2nd paragraph to the very first intro section, something along the lines:**

The SPC data structure is very general. In the SPC data structure, octrees provide a way to store and efficiently retrieve coordinates of points at different levels of the octree hierarchy. It is also possible to associate features to these coordinates using point ordering in memory. Below we detail the low-level representations that comprise SPCs and allow corresponding efficient operations. We also provide a convenience container [[INSERT LINK]] for these low-level attributes.

kaolin/render/spc/raytrace.py
tovacinni: Done

kaolin/render/spc/raytrace.py
tovacinni: Done ish

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,316 @@
...
+
+from kaolin import _C
+from .uint8 import bits_to_uint8
+from kaolin.rep import Spc
+
+def points_to_octree(points, level, sorted=False):
+    r"""Convert points to an octree.
+
+    Args:
+        points (torch.ShortTensor):
```
Caenorst: @tovacinni please add some details about what those `points` are, make sure to show the distinction with spc `points`. (they are at a specific level). Add an `unbatched_` prefix. I think there should be a `points = points.contiguous()` at the very beginning of the function?

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,316 @@
...
+        morton = torch.sort(points_to_morton(torch.unique(points, dim=0)))[0]
+        points = morton_to_points(morton)
+    return _C.ops.spc.points_to_octree(points, level)
+
+
+def points_to_morton(points):
+    r"""Convert points to morton codes.
+
+    Args:
+        points (torch.ShortTensor):
```
Caenorst: @tovacinni same remark than for `points_to_octrees`

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,316 @@
...
+    """
+    return _C.ops.spc.spc_point2morton(points.contiguous())
+
+def morton_to_points(morton):
```
Caenorst: `same remark than for above`

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,316 @@
...
+    r"""Calculates the corners of the points assuming each point is the 0th bit corner.
+
+    Given a [N, 3] tensor returns a [N, 8, 3] tensor.
+
+    Args:
+        points (torch.ShortTensor):
+    """
+    return _C.ops.spc.spc_point2corners(points.contiguous())
+
+def points_to_coeffs(x, points):
```
Caenorst: @tovacinni same remark than above

kaolin/ops/spc/spc.py
```diff
@@ -0,0 +1,316 @@
...
+        octree (torch.ByteTensor):
+        points (torch.ShortTensor):
+        pyramid (torch.IntTensor):
+        exsum (torch.IntTensor):
+        query_points (torch.ShortTensor):
+            A collection of query indices.
+        level (int):
+            The level of the octree to query from.
+    """
+    remaining_kwargs = kwargs.keys() - Spc.KEYS
```
Caenorst: We should remove the usage of `Spc.to_dict` from `query` since it's unbatched

kaolin/render/spc/raytrace.py
```diff
@@ -0,0 +1,96 @@
...
+    Raytrace will already get correspondences, but this will additionally compute distances.
+    Likely to be folded into raytrace in a future version.
+
+    Args:
+        nuggets (torch.IntTensor): the ray-point correspondences. [num_nuggets, 2]
+        points (torch.ShortTensor): the points associated to the octree. [num_points, 3]
+        ray_o (torch.FloatTensor): ray origins. [num_rays, 3]
+        ray_d (torch.FloatTensor): ray directions. [num_rays, 3]
+        level (int): level of the SPC to trace. 
+        info (torch.BoolTensor): First hits. Will compute if None.
```
Caenorst: Need shape info

kaolin/render/spc/raytrace.py
```diff
@@ -0,0 +1,96 @@
...
+    Likely to be folded into raytrace in a future version.
+
+    Args:
+        nuggets (torch.IntTensor): the ray-point correspondences. [num_nuggets, 2]
+        points (torch.ShortTensor): the points associated to the octree. [num_points, 3]
+        ray_o (torch.FloatTensor): ray origins. [num_rays, 3]
+        ray_d (torch.FloatTensor): ray directions. [num_rays, 3]
+        level (int): level of the SPC to trace. 
+        info (torch.BoolTensor): First hits. Will compute if None. 
+        info_idxes (torch.IntTensor): Packed indices of first hits. Will compute if None.
```
Caenorst: Need shape info.

kaolin/render/spc/raytrace.py
```diff
@@ -0,0 +1,96 @@
...
+        mask (torch.BoolTensor): Mask to determine if the ray is still alive. Default to false if None.
```
Caenorst: Need shape info.

kaolin/render/spc/raytrace.py
```diff
@@ -0,0 +1,96 @@
...
+        init (bool): First run? Enables slight optimizations.
```
Caenorst: Need more details. First run of what? if it optimize why is it not always on?

kaolin/render/spc/raytrace.py
```diff
@@ -0,0 +1,96 @@
...
+    Returns:
+        first_hits (torch.BoolTensor): the boolean mask marking the first hit by depth.
+    """
+    return _C.render.spc.mark_first_hit(nuggets.contiguous())
+
+def ray_aabb(nuggets, points, ray_o, ray_d, level, info=None, info_idxes=None, mask=None, init=True):
+    r"""Ray AABB intersection with points. 
+
+    Raytrace will already get correspondences, but this will additionally compute distances.
+    Likely to be folded into raytrace in a future version.
```
Caenorst: you can add a note like that:
```
.. note::

  This function to be folded into raytrace in a future version.

Args:
```

Issue #401, opened 2021-06-08
Sparse adjacency matrix and laplacian matrix problem


AndresCasado:
There is an inconsistency in the `adjacency_matrix` function in `kaolin.ops.mesh` that is making `uniform_laplacian` return a wrong matrix.

When `adjacency_matrix` has the parameter `sparse=True` (default behaviour), it accounts the shared edges twice. In the following mesh, it counts the edge 0-3 twice. See the code below.
```
0-----1
|\    |
| \   |
|  \  |
|   \ |
|    \|
2-----3
```

Code sample:
```python
import torch
from kaolin.ops.mesh import adjacency_matrix
f = torch.tensor([[0, 1, 3], [3, 2, 0]])

print('Sparse to dense')
print(adjacency_matrix(4, f).to_dense())
print()

print('Dense')
print(adjacency_matrix(4, f, sparse=False))
print()

```

Output:
```
Sparse to dense
tensor([[0., 1., 1., 2.],
        [1., 0., 0., 1.],
        [1., 0., 0., 1.],
        [2., 1., 1., 0.]])

Dense
tensor([[0., 1., 1., 1.],
        [1., 0., 0., 1.],
        [1., 0., 0., 1.],
        [1., 1., 1., 0.]])
```

Instead of setting `sparse = False`, the function `uniform_laplacian` calls `adjacency_matrix` and then calls `to_dense()` on the result, which, as shown above, is not correct. https://github.com/NVIDIAGameWorks/kaolin/blob/4f01dda27ac1655b189918c323473d4b603dcece/kaolin/ops/mesh/mesh.py#L104

# 

The fastest fix for `uniform_laplacian` would be to use `sparse=False` directly.
```python
dense_adjacency = adjacency_matrix(num_vertices, faces, sparse=False)
```

# 

For `adjacency_matrix` the solution would be to avoid the repetition of edges. For example, calling `unique` along dimension 0 before transposing the indices seems to work, although I don't know how efficient it is on large matrices.

Original code:
https://github.com/NVIDIAGameWorks/kaolin/blob/4f01dda27ac1655b189918c323473d4b603dcece/kaolin/ops/mesh/mesh.py#L70-L74

Proposed edit:
```python
    if sparse:
        indices = indices.unique(dim=0)
        indices = indices.t()
        # If vertex i and j have an edge connect to it, A[i, j] = 1
        values = torch.ones(indices.shape[1], device=device)
[...]

PR #402, opened 2021-06-08
Fix sparse adjacency matrix by removing duplicates

AndresCasado:
Simple fix for issue #401 

While it works, I should note that there seems to be a problem with the efficiency of the `unique` method. See the following issue in PyTorch repo: https://github.com/pytorch/pytorch/issues/15804

I hope it's ok to reference another projects: PyTorch3D faced this problem and solved it with a hash.

https://github.com/facebookresearch/pytorch3d/blob/f00ef66727194df34492ecbc75d81006c96fb424/pytorch3d/structures/meshes.py#L1076-L1083

### DIFF
--- a/tests/python/kaolin/ops/test_mesh.py
+++ b/tests/python/kaolin/ops/test_mesh.py
@@ -14,6 +14,7 @@
 
 import math
 import pytest
+import os
 
 import torch
 
@@ -22,6 +23,9 @@ from kaolin.ops.batch import get_first_idx, tile_to_packed, list_to_packed
 from kaolin.utils.testing import FLOAT_TYPES, with_seed, check_tensor
 from kaolin.ops import mesh
 from kaolin.ops.mesh.trianglemesh import _unbatched_subdivide_vertices
+from kaolin.io import obj
+
+ROOT_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), '../../../samples/')
 
 @pytest.mark.parametrize("device,dtype", FLOAT_TYPES)
 class TestFaceAreas:
@@ -341,7 +345,7 @@ def test_adjacency_matrix_dense(device, dtype):
 
 @pytest.mark.parametrize('device, dtype', FLOAT_TYPES)
 def test_adjacency_consistent(device, dtype):
-    test_mesh = obj.import_mesh('tests/samples/model.obj')
+    test_mesh = obj.import_mesh(os.path.join(ROOT_DIR, 'model.obj'))
     vertices = test_mesh.vertices
     faces = test_mesh.faces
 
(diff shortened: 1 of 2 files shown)


Caenorst:
Hi @AndresCasado !

Thank you for your quick PR.

I think you forgot to sign your commit, you can follow the DCO suggestion:
```
git commit --amend --no-edit --signoff
```

The issue that you referenced is specifically talking about `torch.unique` with `return_inverse=True` which is not the case here, so hopefully it should be alright.

We maybe need to use `torch.unique` in this setting in the future so that could be interesting to optimize, we would be eager to see a use case where `torch.unique` have a non-negligible slowdown in a training or inference.

Caenorst:
DCO issue is solved.

Looks like the example in the docstring is wrong now that the output order is changed by `torch.unique`.

Also to track regression can you add a test for comparing sparse.to_dense() and dense with adjacency from the mesh `tests/samples/model.obj` ?

If you are unfamiliar with our tests I can guide you.

AndresCasado:
I have just tested the performance of `unique` removing the duplicated vertices of a big mesh (>500k vertices). On GPU it runs in just 1 second, while on CPU it runs in more than 20. Anyways, considering that the common use case is to run it on GPU, I think it shouldn't be a problem.

I've updated the docstring.

About the tests, I guess it should be something similar to this https://github.com/NVIDIAGameWorks/kaolin/blob/4f01dda27ac1655b189918c323473d4b603dcece/tests/python/kaolin/ops/test_mesh.py#L313-L326 but I don't know how to do it with the mesh you mentioned.

Caenorst:
You would import the `tests/samples/model.obj` as in https://github.com/NVIDIAGameWorks/kaolin/blob/385b17699e2c78cad3b18b28c1cba5e80d8fe349/tests/python/kaolin/io/test_obj.py#L121 . and apply the adjacency function on the `faces`.

ps: you can ignore the CI error.

AndresCasado:
Hi, I pushed a new commit with a test

tests/python/kaolin/ops/test_mesh.py
```diff
@@ -339,6 +339,20 @@ def test_adjacency_matrix_dense(device, dtype):
                              [1, 1, 0, 0, 0]], dtype=torch.float, device=device)
     assert torch.equal(output, expected)
 
+@pytest.mark.parametrize('device, dtype', FLOAT_TYPES)
+def test_adjacency_consistent(device, dtype):
+    test_mesh = obj.import_mesh('tests/samples/model.obj')
```
Caenorst: `obj` isn't been imported.

We recommend avoiding using relative path in case test are launched from a different place, you can use the following approach: https://github.com/NVIDIAGameWorks/kaolin/blob/master/tests/python/kaolin/io/test_obj.py#L22

Caenorst (commented):
The new test is not passing, but a small fix should do.

If you want to execute the test on your side:
1) install by running: `pip install -r tools/ci_requirements.txt`
2) execute `pytest tests/python/kaolin/ops/test_mesh.py`

AndresCasado:
Hi

I've fixed both issues. The test is passing now and I've set up the directory like the example you sent me.

Caenorst (approved):
LGTM! Thank you for the contribution! :rocket:

PR #421, opened 2021-08-09
Accelerate USD import.

mlej8:
Accelerate USD import > 100x, profiling import of 100 random USD assets goes from 646s to 2.4s.

### DIFF
--- a/tests/python/kaolin/io/test_usd.py
+++ b/tests/python/kaolin/io/test_usd.py
@@ -121,6 +121,11 @@ class TestMeshes:
         with pytest.raises(usd.NonHomogeneousMeshError):
             usd.import_meshes(hetero_mesh_path, ['/Root'])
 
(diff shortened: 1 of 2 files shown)


tests/python/kaolin/io/test_usd.py
```diff
@@ -121,6 +121,11 @@ def test_import_hetero_fail(self, scene_paths, out_dir, hetero_mesh_path):
         with pytest.raises(usd.NonHomogeneousMeshError):
             usd.import_meshes(hetero_mesh_path, ['/Root'])
 
+    def test_import_hetero_fail2(self, scene_paths, out_dir, hetero_mesh_path):
```
Caenorst: Can we have more explicit naming for the tests ? (rather the suffix `2` which is not expressing what is the difference with the first version)

mlej8: Update test names.

Caenorst (approved):
LGTM ! Thanks ! :rocket:

PR #424, opened 2021-08-16
Adding exporting pointcloud as point instancer.

JerryJiehanWang:


Caenorst (approved):
Approving as the CI is stuck, tested on local workstation

PR #434, opened 2021-08-27
Fix issue with contiguous inside points to octree

tovacinni:
The previous implementation without contiguous will occasionally cause issues (the created octree being wrong). This is a small patch to fix that issue.


Caenorst (approved):
LGTM! thanks for the fix!

PR #437, opened 2021-09-01
Fix illegal memory access bug in raytrace

tovacinni:
In `d_Decide` in raytrace, there is an octree lookup that is performed to determine the number of voxels to allocate for the next level of the raytrace. This however results in a illegal memory access when tracing the last level of the octree. The simple solution is to only do this operation if it is determined that the raytrace is not at the last level yet.


Caenorst (approved):
LGTM

PR #456, opened 2021-10-25
Fix #448, Added a new mode for extract_surface

### LINKED ISSUE #448: Surface extraction in voxels is wide
Hello, 

In kaolin/ops/voxelgrid.py, extract_surface() removes internal voxels to keep only external voxels. It does this by keeping the filled voxels that are next to an empty voxel in a 3x3 radius. In the Trimesh library, there is a hollow function that does the same on the processor, but it has an option to choose the adjacency matrix, and by default it only keeps filled voxels that are directly next to a empty voxel. It gives a finer surface.

I made an illustration in 2D, but it is the same in 3D : 

```
import matplotlib.pyplot as plt
import torch
import numpy as np

device = "cuda" if torch.cuda.is_available() else "cpu"

noise = np.random.randn(3, 3) + 1j * np.random.randn(3, 3)
binary_map = np.fft.ifft2(noise, (100, 100)).real
binary_map = binary_map > np.median(binary_map)

plt.imshow(binary_map)
plt.show()
```
First i made a binary image made of random blob (yellow=True)
Figure_1

```
def trimesh_surface(dense, structure=None):
    from scipy.ndimage import binary_dilation
    # dense = np.pad(dense, np.ones((3, 2), dtype=int), mode='constant')
    dense = np.pad(dense, np.ones((2, 2), dtype=int), mode='constant')
    empty = np.logical_not(dense)
    dilated = binary_dilation(empty, structure=structure)
    # surface = np.logical_and(dense, dilated)[1:-1, 1:-1, 1:-1]
    surface = np.logical_and(dense, dilated)[1:-1, 1:-1]
    return surface


trimesh_map = trimesh_surface(binary_map)
plt.imshow(trimesh_map)
plt.show()
```
Thin surface
Figure_2

```
def kaolin_surface(voxelgrids):
    import torch.nn.functional as F

    voxelgrids = voxelgrids.float()

    # output = F.avg_pool3d(voxelgrids.unsqueeze(1), kernel_size=(3, 3, 3), padding=1, stride=1).squeeze(1)
    output = F.avg_pool2d(voxelgrids.unsqueeze(1), kernel_size=(3, 3), padding=1, stride=1).squeeze(1)

    output = (output < 1) * voxelgrids.bool()

    return output


kaolin_map = torch.from_numpy(binary_map).unsqueeze(0).to(device)
kaolin_map = kaolin_surface(kaolin_map)
kaolin_map = kaolin_map.squeeze(0).cpu()
plt.imshow(kaolin_map)
plt.show()
```
Wide surface
Figure_3

I want to extract the surfaces of the voxels and then calculate the SDF, and I guess I will have better results if I use the thinnest layer. Is there a reason why the wide surfaces is preferred? I could do a PR to make the two options available (wide and thin), would that be useful ?

le-Greg:
Hello,
Fix #448. I also deleted one test (test_output_batch) that was redundant with the previous one (test_output_value), and replaced a try...except by a if...raise because of code readability. There was also a comment that I think was outdated, because the function actually handles boolean

### DIFF
--- a/kaolin/ops/voxelgrid.py
+++ b/kaolin/ops/voxelgrid.py
@@ -87,12 +87,16 @@ def downsample(voxelgrids, scale):
     return output.squeeze(1)
 
 
-def extract_surface(voxelgrids):
+def extract_surface(voxelgrids, mode="wide"):
     r"""Removes any internal structure(s) from a voxelgrids.
 
     Args:
         voxelgrids (torch.Tensor): Binary voxelgrids of shape (N, X, Y ,Z)
                                    from which to extract surface
+        mode (str): Either "wide" or "thin". Each voxel can be seen as a cube in a grid.
+                    "wide" mode keeps each filled voxel with at least one vertex in contact
+                    with an empty voxel. "thin" mode keeps each filled voxel with at least
+                    one face in contact with an empty voxel.
 
     Returns:
         torch.BoolTensor: binary surface voxelgrids tensor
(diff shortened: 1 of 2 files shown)


Caenorst (approved):
LGTM thanks for the contributions!

PR #471, opened 2021-11-16
add macro for error message when kaolin not built with CUDA

Caenorst:
Signed-off-by: Clement Fuji Tsang <cfujitsang@nvidia.com>


tovacinni (approved):
Looks good to me!

PR #476, opened 2021-11-23
adding marching tetrahedra

frankshen07:
Adding the differentiable marching tetrahedra function.


Caenorst:
you want to forward `marching_tetrahedra` in `kaolin/ops/conversions/__init__.py` as all the other conversions.

Also add `kaolin/ops/conversion/tetmesh.py` to https://github.com/NVIDIAGameWorks/kaolin/blob/master/docs/kaolin_ext.py#L35

Caenorst (approved):
LGTM! Thanks for the first tetrahedral feature! :D

Issue #463, opened 2021-11-04
Depth from SPC Raytrace


tovacinni:
The current implementation for SPC raytracing computes depth as part of the AABB mechanism (which uses Majercik et al https://jcgt.org/published/0007/03/04/), but does not return the depth (and instead another kernel recomputes depth, which is wasteful). 

We should instead return the entry (and optionally exit) depths to each voxel from the ray AABB intersections.

This should also include some UX improvements, like:

- Return a tuple of indices instead of a joint tensor
- Change mark_first_hits to take indices

This is currently being addressed in an internal MR (126).

PR #478, opened 2021-12-01
Add depth outputs from SPC raytracing, slight API refactor, fix raytrace AABB behaviour

tovacinni:
Summary of changes: 

- `mark_first_hits` is now `mark_pack_boundary` which a more general packed operation.
- SPC `raytrace` can now return depth. This deprecates the need for the old `ray_aabb` function. 
- Some edge cases in raytracing is handled differently now. Before, rays could miss if they are exactly between 2 voxels, but now it will hit both. 
- `raytrace` will now return 2 tensors, the ray index and the point hierarchy index. Previously these 2 tensors were stored as a single tensor, which was confusing.
- Other misc enhancements and fixes. 

This also addresses https://github.com/NVIDIAGameWorks/kaolin/issues/463 .


Caenorst (approved):
LGTM! thanks!

PR #526, opened 2022-02-03
loosen test precision tolerance to pass on RTX 3090 on windows

zookae:
Testing on RTX 3090 on windows fails with the current precision tolerance values. This MR loosens the minimal subset of tolerances to pass. Covers `test_conv.py` and `test_gcn.py` (adding new explicit tolerance thresholds).


Caenorst (approved):
LGTM!

PR #543, opened 2022-03-31
Example - fitting a 3d bounding box using differentiation

zookae:
Example of using DIBR in a simple optimization loop to fit a 3d bounding box given a set of 2d renders with camera extrinsics.
Adds one new object file and the example notebook.

The notebook demonstrates setting up differentiable rendering and connecting the gradients to parameters of a 3d bounding box. Execution trains for a few epochs, periodically renders test images, and generates a final animation of the training process. Training differentiates through parameters of a 3d bounding box representing the object rotation, translation, and scale (rigid transforms).

Signed-off-by: Alex Zook <azook@nvidia.com>


Caenorst (approved):
LGTM, thanks!

PR #585, opened 2022-06-20
fix: use elif in the displacement conditional

ElijahPepe:
In materials.py, the displacement check doesn't utilize name.lower() nor the rest of the branch logic in the _read_data function. Based on the rest of the if statements in the _read_data function, this is probably a mistake.


shumash:

PR #622, opened 2022-09-09
Fix 404 links in tutorial_index.rst

Mason-McGough:
Fixes several broken links in the Camera section of tutorial_index.rst.


Caenorst (approved):
LGTM, good catch and thank you for the fix!

PR #726, opened 2023-06-04
Update camera_ray_tracing.py: fix small bugs

YuQiao0303:
Fix the following 2 bugs:
1. 'CameraFOV' cannot be imported from kaolin.render.camera, but should be from kaolin.render.camera.intrinstics instead.
2. 'width' and 'height' of the generated pixel grid and camera should be consistent, otherwise the average of generated ray_dir does not equal to the desired look_at direction.

### DIFF
--- a/examples/recipes/camera/camera_ray_tracing.py
+++ b/examples/recipes/camera/camera_ray_tracing.py
@@ -6,7 +6,8 @@
 import torch
 import numpy as np
 from typing import Tuple
-from kaolin.render.camera import Camera, CameraFOV
+from kaolin.render.camera import Camera
+from kaolin.render.camera.intrinsics import CameraFOV
 
 def generate_pixel_grid(res_x=None, res_y=None, device='cuda'):
     h_coords = torch.arange(res_x, device=device)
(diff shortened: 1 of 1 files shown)


examples/recipes/camera/camera_ray_tracing.py
```diff
@@ -6,7 +6,8 @@
 import torch
 import numpy as np
 from typing import Tuple
-from kaolin.render.camera import Camera, CameraFOV
+from kaolin.render.camera import Camera
+from kaolin.render.camera.intrinsics import CameraFOV
```
orperel, line 10: This one isn't actually needed as `__init__.py` has: 
```python3
from .intrinsics import *
```

examples/recipes/camera/camera_ray_tracing.py
```diff
@@ -45,20 +46,21 @@ def generate_perspective_rays(camera: Camera, pixel_grid: Tuple[torch.Tensor, to
 
     return ray_orig, ray_dir, camera.near, camera.far
 
-
+img_width = 200
+img_height =200
```
orperel, line 50: Minor: Please fix the space here

examples/recipes/camera/camera_ray_tracing.py
```diff
@@ -45,20 +46,21 @@ def generate_perspective_rays(camera: Camera, pixel_grid: Tuple[torch.Tensor, to
...
     x0=0.0, y0=0.0,
-    width=800, height=800,
+    width=img_width, height=img_height,
     near=1e-2, far=1e2,
     dtype=torch.float64,
     device='cuda'
 )
 
-pixel_grid = generate_pixel_grid(200, 200)
+pixel_grid = generate_pixel_grid(img_width, img_height)
```
orperel, line 63: Given that this snippet should be a simple recipe, I agree with your suggestion of having the raygen grid and image plane with the same resolution.
To complete the picture however - the raygen grid resolution and image plane resolution can actually differ, and this can also be exploited to, i.e., trace a lower-resolution image.

For reference, kaolin-wisp implements exactly that for its interactive mode:
1. https://github.com/NVIDIAGameWorks/kaolin-wisp/blob/main/wisp/renderer/core/render_core.py#L167C9-L167C9
2. https://github.com/NVIDIAGameWorks/kaolin-wisp/blob/main/wisp/renderer/core/render_core.py#L290

Picking an arbitrary resolution (which is not a power of 2 downscale factor) can indeed cause alignment issues, which I think it what you're worried about?

orperel (changes requested):

PR #760, opened 2023-10-11
omitting R G B from .obj vertex data

mikerjacobi:
Wavefront .obj files support `v X Y Z R G B`, which get interpreted by `kaolin/io/obj.py:import_mesh` as additional vertices. This change omits the R G B values. 

Addresses the same issue as [this PR](https://github.com/NVIDIAGameWorks/kaolin/pull/300), but in a different way.


Caenorst (approved):
LGTM! Thanks for the MR!

PR #765, opened 2023-11-07
support torch 2.1.0 and cuda 12.1

Caenorst:


samiede:
Is this something that is going to be included in a release soon? It is blocking for us and direly needed!

PR #881, opened 2025-03-29
Support Blackwell

johnnynunez:


johnnynunez:
@Caenorst can you merge it? It is for blackwell (5090)

Caenorst:
Hi @johnnynunez , you can just build with environment variable `IGNORE_TORCH_VER=1`, I wouldn't explicitely put 2.8.0 is supported from Kaolin because that's not something I've thoroughly tested (especially since those are nightly builds and might create breaking changes)

PR #921, opened 2025-08-23
Fix: Typo in FlexiCubes leading to program crash

Maxmilite:
This PR fixes a typo in the FlexiCubes module that was causing a program crash.

Specifically, the function call `ckaolin.utils.testing.heck_tensor` has been corrected to `kaolin.utils.testing.check_tensor`.


Caenorst (approved):
Hi @Maxmilite , thank you so much for the PR! looks like we introduced a bug as we stop tracking the the non commercial version of flexicube. In the meanwhile we invite most users to use the Apache V2.0 version so their code can be used commercially :)

PR #927, opened 2025-09-22
Speed up marching_tets

JulianKnodt:
Using raw indexing is slower than using index_select, and flip also creates an unnecessary copy. I found that this speeds up marching_tetrahedra by a small margin, so thought it was worth merging upstream.

### DIFF
--- a/kaolin/ops/conversions/tetmesh.py
+++ b/kaolin/ops/conversions/tetmesh.py
@@ -78,15 +78,18 @@ def _unbatched_marching_tetrahedra(vertices, tets, sdf, return_tet_idx):
         mapping[mask_edges] = torch.arange(mask_edges.sum(), dtype=torch.long, device=device)
         idx_map = mapping[idx_map]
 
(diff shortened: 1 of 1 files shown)


Caenorst:
Hi @JulianKnodt , can you rebase? there were some issue with the doc build that got fixed recently. Also do you have a case where the speedup is obvious? I tried with our tutorial and the speedup seems quite negligeable.

JulianKnodt:
I'm currently using marching tets for a research project that is not yet released, and I no longer have access to the machine (/ any other gpus) I was using, but I found that at a high enough resolution it was faster. I didn't try the tutorial, but I was using it with differentiable rendering so maybe you'll see a speed up with flexicubes.

Caenorst:
LGTM! I was able to profile the specific speedup and see it in the backward. In the full pipeline I couldn't really see the speedup so I assume this depend on hyperparameters. Thanks @JulianKnodt .

PR #992, opened 2026-07-10
Fix equivolume batch broadcasting

ShiroKSH:
## Summary

- reshape per-mesh tetrahedron means along the batch axis
- add a regression test with different batch and tetrahedron counts
- correct the documented output shape and example values

## Root cause

`equivolume` reduced tetrahedron volumes to a tensor of shape `(B,)`, then reshaped it to `(1, B)` before subtracting it from volumes shaped `(B, T)`. This raised a broadcasting error when `B != T` and mixed means between meshes when `B == T`.

The mean now has shape `(B, 1)`, so each mesh is compared with its own mean volume.

## Validation

- all four `TestTetMeshMetrics` checks pass in an isolated CPU PyTorch environment
- batched output matches per-mesh evaluation
- backward pass produces finite gradients; gradcheck passes
- flake8 and `py_compile` pass for the changed files

### DIFF
--- a/kaolin/metrics/tetmesh.py
+++ b/kaolin/metrics/tetmesh.py
@@ -61,7 +61,7 @@ def equivolume(tet_vertices, tetrahedrons_mean=None, pow=4):
         tetrahedrons_mean (torch.Tensor):
             Mean volume of all tetrahedrons in a grid,
             of shape :math:`(\text{batch_size})` or :math:`(1,)` (broadcasting).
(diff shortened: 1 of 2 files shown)


ShiroKSH:
@Caenorst - could you take a look at the equivolume batch-broadcasting fix in this PR when you have a chance, or route it to the right reviewer? I can make any requested changes or add tests. Thanks!

ShiroKSH:
Hi maintainers, this PR has been open for over two weeks. When you have a moment, I would appreciate a review. Thank you.

PR #1000, opened 2026-08-14
small fixes for farthest point sampling 

nmwsharp:
This PR fixes three minor issues with the farthest point sampling algorithm:

1. Adds a proper check with error if the user attempts to invalidly sample `k > N` points. (Old behavior: out of bounds, or silent hang)
2. Fix a wrong sign comparison, causing padding to never be actually applied.
3. Fix a bad short-circuit operation in Warp, which was causing an out of bounds access

2 & 3 were both correctness issues; interestingly we never saw them cause a problem, but this was just luck as they were definitely invalid access. Both were correctly caught when running with `warp.config.mode = "debug"`, which is now clean.

### DIFF
--- a/kaolin/ops/pointcloud.py
+++ b/kaolin/ops/pointcloud.py
@@ -506,12 +506,23 @@ def _find_farthest_point_ind_kernel(
     my_dist = distancesSq[i]
 
     curr_farthest_ind = farthest_point_inds[i_round]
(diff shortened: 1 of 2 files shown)


nmwsharp:
As an aside, we may want to enable `warp.config.mode = "debug"` for our tests, to catch such issues earlier in the future. But for now this PR does not yet do so.

